PR #1, opened 2012-12-12
Mezzanine fixes from rebol-patches

BrianHawley:
Ported over some of the fixes from the rebol-patches project.


carls:
Thanks Brian. Good catch on the file endings. I tend not to see those with my default editor settings.

BrianHawley:
I also have editors which support LF. In this case, however, Github for Windows converted the line endings on clone, and apparently expected your project to have this setting to convert them back on commit. Without the setting, the commits stayed as CRLF.

This will make it so more people can just use the GUI tools and put off learning the git command line. It's good to do what we can to avoid discouraging initial adoption.

PR #2, opened 2012-12-12
Fix build: rename obj/ to objs/

earl:
The makefile currently depends on an objs/ directory to exist, but we
ship an obj/ directory. Fix the immediate issue by renaming obj/ to
objs/. Adapted the README accordingly.

Ultimately, (a) obj/ probably is the better name and (b) the build
should create it if it does not already exist.


earl:
Obsoleted by 8657fbcff9465ac7d09a800333786d4e15343548.

PR #4, opened 2012-12-13
Fix mold/all of decimals on Linux (CC1633)

earl:
The results of mold/all for decimals on Linux are inconsistent with other platforms (Win32, OSX).

For example, on R3 2.101.0.4.4 (Linux):

```
>> mold/all 0.1
== "0.1"
```

Whereas on R3 2.100.111.3.1 (Win32):

```
>> mold/all 0.1
== "0.10000000000000001"
```

See [CureCode #1633](http://issue.cc/r3/1633)  for details.

As suggested by Ladislav Mecir in the CureCode discussion, undefining the `HAS_ECVT` config macro fixes this issue.

### DIFF
--- a/src/include/reb-config.h
+++ b/src/include/reb-config.h
@@ -140,7 +140,6 @@ These are now obsolete (as of A107) and should be removed:
 
 #ifdef TO_LINUX					// Linux/Intel
 #define ENDIAN_LITTLE
-#define HAS_ECVT
 #define HAS_LONG_DOUBLE
 #endif
 


earl:
Well, `mold/all 0.1` working comes at the expense of a regression. Without `HAS_ECVT` the following is false in Linux builds:

```
same? 9.9999999999999926e154% load mold/all 9.9999999999999926e154%
```

Whereas with `HAS_ECVT` for Linux the above is true as it already is on Win32, OSX.

carls:
Yes, we've spent a lot of time on it. It's been non-trivial because of the side effects... and also that the code has to be tested on various systems which may differ in small ways.

Ask around and see if you can get a few others to help figure out the best fix to make and get it tested. Ladislav is really good with this stuff too, so perhaps ask him for advice.

carls:
Earl, can you figure out how to fix this without the regression? Maybe ask Ladislav for help?

earl:
I've already been briefly talking to Ladislav about this, yes. I'm looking into the build-related fixes now, will get back to the decimals fix later. If you prefer to not have the open pull request around, please do not hesitate to close it and I'll redo when/if we could come up with a more thorough fix.

PR #7, opened 2012-12-14
Fixes to some critical tickets, and some docs

BrianHawley:
Lots!

Simple fixes to a couple native problems I already knew how to fix, or could figure out easily.

A critical workaround for a problem with `resolve/extend/only` that no-one has been able to diagnose properly yet. It's critical because it makes R3's module system crash R3 at inexplicable times, and has led to some people dropping R3 altogether. Left the old code in there commented out, for when we fix `resolve/extend/only`.

A couple critical comment updates, one to better explain the use of the `sys/load-header function`, and another to explain an ambiguous comment in the startup code that would prevent bugs from being readded.

Lastly, making the rest of the loading functions support the length-specified embedded scripts that `sys/load-header` supported already. While I was at it, `sys/load-module` and `sys/load-ext-module` now can be used to make a multi-script-per-file loader.

### DIFF
--- a/README
+++ b/README
@@ -67,3 +67,10 @@ Enjoy your new REBOL freedom,
 
 -Carl
 
+Bug reporting note:
+
+Report bugs, wishes, etc. at the
+
+    http://curecode.org/rebol3/view-tickets.rsp
+
+site.


BrianHawley:
Added a standard .gitignore so our repos don't get cluttered with operating system metadata and build process artifacts from Eclipse, Visual Studio, or various installer builders.

carls:
Brian, is it possible on github to request separate pulls for simple low risk changes from other code changes? Many of the simple ones like the .gitignore, comments/docs, typos can be merged without rebuilding and testing R3.

earl:
Not possible to do "after the fact" via the Github web interface. But it is possible to do clean this up with Git: push unrelated changes to separate branches, and open a pull request for each. That's generally the preferred way to do it, to make both reviewing and integration of changes easier.

BrianHawley:
Sorry, I'll do that going forward.

Most of these are pretty trivial, the documentation fixes, the .gitignore thing, the native fixes. The really critical thing is that sys-load.r revamp - we've been waiting for that fix for a really long time, and R3 is really crash-prone without it. I even had to figure out how to patch R3 functions in order to fix it from the outside earlier this year, and without it modules are useless. Also, it finishes implementing that length-specified embedded script feature that you had me add a couple years ago. The sys-load.r fixes are ported from my patches collection, which I have been running with for many months now, though I only published recently.

I'm pretty much done with the quick fixes, as far as I know. Maybe there will be more once I've reviewed more of the code. Now I'm trying to learn enough of the code base to work on the big fixes.

earl:
Brian, I'll get in touch re splitting up the above into smaller pieces.

BrianHawley:
OK, it you need to throw away the minor changes, we can redo them later. The only critical ASAP changes are the sys-load.r ones, which require recompiling/retesting R3. If they can be applied right away, we can move forward. I am blocked until you can accept the sys-load.r changes (and possibly the .gitignore one too since I would be using VS, not MinGW).

I was serious about the _going forward_ part. If there is a way to split up the pull, that's nice, and I'd be able to do that today (Earl?). If you have to focus on one change at a time, start with the sys-load.r stuff first (which I've already tested), then the native fixes.

PR #13, opened 2012-12-14
Add the -m32 option for the Linux x64 target to the linker-flags.

spytheman:
The -m32 option should be added in the linker-flags too, 
or else building on linux x64 fails with errors  like these:

```
/usr/bin/ld: i386 architecture of input file `objs/a-constants.o' is incompatible with i386:x86-64 output
```

Adding it fixes the build process on Linux x64 (Ubuntu 12.04) .


earl:
Thanks. This is now obsoleted by 0ffda62e088b1d3cfc383a0475c7cab459fd7184, I guess.

spytheman:
No, it is not obsolete. This commit is an _addition_ to 0ffda62 . 
When linking the final executable, the -m32 option should be passed too, or else the build fails on a 64bit linux.

earl:
Right, the link flag is still missing even with 0ffda62. Tested the proposed fix with GCC 4.7.2 x86_64-unknown-linux-gnu , works fine.

PR #22, opened 2012-12-14
mingw32 GCC errors out when building R3 core lib

cyphre:
I made a fix to the GCC error when building under Windows:

```
gcc ../src/core/s-make.c -c -DTO_WIN32 -DREB_API -DUNICODE -O2 -I. -I../src/incl
ude/ -o objs/s-make.o
../src/core/s-make.c: In function 'Copy_OS_Str':
../src/core/s-make.c:133: error: lvalue required as increment operand
../src/core/s-make.c:141: error: lvalue required as increment operand
```

GCC version I'm using is:

```
C:\Users\cyphre>gcc -v
Using built-in specs.
Target: mingw32
Configured with: ../../gcc-4.4.1/configure --prefix=/mingw --build=mingw32 --ena
ble-languages=c,ada,c++,fortran,objc,obj-c++ --disable-nls --disable-win32-regis
try --enable-libgomp --enable-cxx-flags='-fno-function-sections -fno-data-sectio
ns' --disable-werror --enable-threads --disable-symvers --enable-version-specifi
c-runtime-libs --enable-fully-dynamic-string --with-pkgversion='TDM-2 mingw32' -
-enable-sjlj-exceptions --with-bugurl=http://www.tdragon.net/recentgcc/bugs.php
Thread model: win32
gcc version 4.4.1 (TDM-2 mingw32)
```

Please review the fix and if it looks ok to you, include in the main repository.


earl:
I had the same error (and did an identical fix) with GCC 4.7.

carls:
Thanks Richard. Let me figure out how to merge this.

carls:
To avoid a hand merge, I need you to rename makefile to makefile.mingw to avoid a conflict.

The makefiles are only starting points for new users. We can include more than one so new users don't need to figure out where to start. The true changes are in the tools files you submitted.

Let me know when you have it ready, and I will merge it.

carls:
BTW, this shows a bit of a flaw in github. It appears I cannot pull just part of the request. I have to pull all or nothing.

For this case, it would be enough to pull all changes except the makefile, then add a new makefile.mingw manually.

Anyway, I need you to make the change and resubmit it. Thanks.

cyphre:
Carl, Andreas is working on make-make.r changes so the mingw makefile will be fully generated using the 0.3.1 arg. So there is no need to have separate makefile.mingw. Andreas will surely post update here once all the changes are ready.

Also  you can even now simply merge the first https://github.com/cyphre/r3/commit/e67d0c7150007085fd80e3f7037907b36868723c commit in this request as it is related only to core/s-make.c file and the C code change is IMO not platform specific.

BrianHawley:
Cyphre, Andreas, please take into account the Windows users who won't be using MinGW. Don't tie the Windows platform too thoroughly to that compiler. Nonetheless, carry on, great work!

cyphre:
Brian, yes we already do. The C code change is compiler independent. Nothing special. Regarding the makefile generation we already keep in mind mingw32 is just one Windows alternative so anyone who is using VCPP or whatever else can just enhance the make-make.r and/or systems.r accordingly

carls:
Cyphre, I agree that I can merge the C changes. But, I often do the pull from github.com web page - just with the click of a button. But, if there is a conflict, it takes more effort. There does not appear to be a way from github.com to only accept specific files in the pull request.

If you can remove the makefile from the pull request, I can merge the rest of the files.

BTW, I am not an expert on git. So, if you have advice, please let me know. Thanks.

earl:
I'll close this pull request and open two separate fresh ones: one for Cyphre's code fix, one for the Win32 build fixes.

PR #28, opened 2012-12-16
Fix and improve build for Win32

earl:
Here's the first set of changes to the Win32 build, created in collaboration with Cyphre and Robert. A quick overview (without trying to reiterate the commit summaries too much):
- Disable `declspec(thread)` for MinGW
- Default to build with REB_EXE defined
- Drop `-lc` linking to libc (not needed on Linux, OSX; MinGW on Win32 barfs on this)
- Link against wsock32 and comdlg32
- Build sources from os/win32/ for Win32
- Build r3 as GUI binary (just like the official RT Alpha builds)
- Fix stack size allocated for the Win32 binaries
- Use .exe in name of Win32 executables
- Use DIR instead of LS on Win32 (makes us slightly more independent from the MinGW shell)
- Use r3-make with `>NUL:`, to avoid R3 crash in make on Win7

The commit messages have all the detail. Be sure to not miss the "files changed" tab, to get a full diff for all the changes (or just [append `.diff` to the pull request URL](https://github.com/rebol/r3/pull/28.diff) to get it as plain text).

### DIFF
--- a/src/include/reb-config.h
+++ b/src/include/reb-config.h
@@ -102,7 +102,9 @@ These are now obsolete (as of A107) and should be removed:
 #define INLINE __inline			// name used for it
 
 #ifdef THREADED
+#ifndef __MINGW32__
 #define THREAD __declspec(thread)
+#endif
 #endif
 
 // Used when we build REBOL as a DLL:
--- a/src/tools/make-make.r
+++ b/src/tools/make-make.r
@@ -63,7 +63,7 @@ I= -I$(INCL) -I$S/include/
 TO_OS=	
 OS_ID=	
 RAPI_FLAGS=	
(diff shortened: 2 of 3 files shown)


carls:
Looks ok.

An alternative to r3$(BIN_SUFFIX) would be to just use $(R3_EXE), where:

R3_EXE = r3$(BIN_SUFFIX)

Same for host, etc.

earl:
Thanks for looking it over and merging.

Yes, we thought about `R3_EXE = r3$(BIN_SUFFIX)` as well: `R3_{CORE,VIEW}{,_HOST}_EXE` and `R3_LIB`. Will do this together with lib suffix cleanup changes, most likely.

PR #30, opened 2012-12-17
Update src/core/a-lib.c

nicolas42:
just a typo


earl:
Thanks. Minor nit: "Fix typo in src/core/a-lib.c" would have been a slightly more self-explaining commit message :)

PR #31, opened 2012-12-17
Bug reporting note added.

ladislav:


ladislav:
Aha, I did not notice you may have already decided to move all CC tickets here instead. (I wrote the note to honor the GrahamC's wish)

PR #34, opened 2012-12-17
fixed PARSE tag bug

cyphre:
This fix should correct PARSE bug reported by Rebolek today.

Here is example of the bug:

```
>>  parse "<tag>string" [thru <tag> m: (probe m)]
"ing"
```

After applying the fix:

```
>>  parse "<tag>string" [thru <tag> m: (probe m)]
"string"
```

Carl, please crosscheck if the fix makes sense.

### DIFF
--- a/src/core/u-parse.c
+++ b/src/core/u-parse.c
@@ -518,10 +518,11 @@ bad_target:
 					i = Find_Str_Str(series, 0, index, series->tail, 1, ser, 0, ser->tail, HAS_CASE(parse));
 					if (i != NOT_FOUND && is_thru) i += ser->tail;
 				}
-				else
+				else {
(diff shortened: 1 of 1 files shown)


BrianHawley:
The fix was to the problem reported in http://issue.cc/r3/682 which was partially fixed in alpha 84.

carls:
Just looking at it, why is this line removed:

524 -    else if (is_thru) i += VAL_LEN(item);

Wouldn't it apply to both conditions in the prior if/else?

cyphre:
Carl the problem was that in case tag! gets parsed this line before:

519 -  if (i != NOT_FOUND && is_thru) i += ser->tail;

already adds the lenght of formed tag

and later on the line 524 the lenght (moreover incorrect) gets added again so the resulting parse index is wrong

cyphre:
So IMO for pure string! and binary! the I variable should be increased by VAL_LEN(item)
but for tag! it should be increased by ser->tail only. (in THRU mode)

PR #35, opened 2012-12-17
Temporary workaround for resolve/extend/only crash (http://issue.cc/r3/1865)

BrianHawley:
RESOLVE/extend/only crashes sometimes, under obsure and (so far) unpredictable circumstances, but unfortunately not rarely. This crash only affects modules which export stuff, and not consistently, but often enough that modules are unusable.

We have to patch this fast. This patch comments out all calls to this function with these options, replacing them with semantically equivalent calls. A little more overhead, but at least it doesn't crash.  We can revert these once someone figures out how to fix the bug in RESOLVE.

### DIFF
--- a/src/mezz/sys-load.r
+++ b/src/mezz/sys-load.r
@@ -67,8 +67,13 @@ export-words: func [
 	words [block! none!] "The exports words block of the module"
 ][
 	if words [
-		resolve/extend/only lib ctx words  ; words already set in lib are not overriden
-		resolve/extend/only system/contexts/user lib words  ; lib, because of above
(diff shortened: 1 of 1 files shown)


earl:
I think I've found the underlying problem crashing RESOLVE/extend/only, so maybe this workaround is no longer necessary. See https://github.com/rebol/r3/pull/41.

BrianHawley:
Please test Andreas' fix before applying my workaround. I would love to not have to do this.

carls:
Checking: can we close this one then?

BrianHawley:
If Andreas' fix works, go ahead and close this one.

PR #40, opened 2012-12-17
Multi-script support, and better support for length-specified embedding

BrianHawley:
The `sys/load-header` function has supported both length-specified and script-in-a-block embedding, but only the latter has been supported by the functions that call it, such as `load`.

This adds support for length-specified embedded scripts to `load`, `load-module`, and `load-ext-module`. It also makes `load-module` and `load-ext-module` also return the end position of their scripts, which makes multi-script loaders possible. It also adds comments to `load-boot-exts` to explain their limitations, and makes sure that the changes to `load-ext-module` don't bother it.

### DIFF
--- a/src/mezz/sys-load.r
+++ b/src/mezz/sys-load.r
@@ -160,6 +160,7 @@ load-header: function/with [
 	]
 	;assert/type [hdr object! rest [binary! block!] end binary!]
 	;assert/type [hdr/checksum [binary! none!] hdr/options [block! none!]]
+	;assert [same? head :source head :end]
 	reduce [hdr rest end]
 ][
 	non-ws: make bitset! [not 1 - 32]
@@ -173,7 +174,7 @@ load-ext-module: function [
 	; for ext obj: help system/standard/extensions
 	assert/type [ext/lib-base handle! ext/lib-boot binary!] ; Just in case
 
-	if word? set [hdr: code:] load-header/required ext/lib-boot [
+	if word? set [hdr: code: end:] load-header/required ext/lib-boot [
 		cause-error 'syntax hdr ext  ; word returned is error code
 	]
 	;assert/type [hdr object! hdr/options [block! none!] code [binary! block!]]
@@ -201,10 +202,15 @@ load-ext-module: function [
 		]
 	]
 
+	; Only decodes the first script if a multi-script
+	if all [not empty? end same? head code head end] [
+		code: to block! copy/part code end  ; Embedded script code
+	]
+
 	; Convert the code to a block if not already:
 	unless block? code [code: to block! code]
 	insert code tmp ; Extension object fields and values must be first!
-	reduce [hdr code] ; ready for make module!
+	reduce [hdr code end] ; copy/part 2 to pass to make module!
 ]
 
 load-boot-exts: function [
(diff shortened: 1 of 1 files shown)


BrianHawley:
Related CureCode ticket here: http://issue.cc/r3/1877 and http://issue.cc/r3/1941

Note that 1941 is a script and module security problem.

carls:
Can you get a few other developers to pull this and test it. It's a more complex change. I don't have time to test it myself. Thanks.

carls:
Any news on test results for this?

Would it makes sense to hold off until next branch

BrianHawley:
I have requested that other developers pull this and test, but they're apparently still catching up and haven't replied yet. I'll try again in more forums. I haven't found any problems with it. If you want to hold off, go ahead, but it might be better in the long run to just go forward.

The length-specified embedding fixes enable a documented feature that doesn't currently work. Once that feature works, the code will basically work like this, as far as externally visible behavior. I have some native enhancements in mind to make this more efficient, but that won't change the API. The multi-script stuff is a side-effect. We already had low-level support for multi-scripts since alpha 109 or earlier. That's why no changes to `sys/load-header` were needed for this fix.

Here are the tickets for the native fixes prompted by this: http://issue.cc/r3/1915 and http://issue.cc/r3/1916

Would it make sense for me to try to remove all externally visible API changes, temporarily, so you can push through the change right away? It would just be a matter of temporarily making `sys/load-ext-module` and `sys/load-module` not return the end position, going back to returning a block of two values instead of three. I can even comment out the existing three-value-return code, so when we add mezzanine-level multi-script support we can restore the changes. I can even make a couple CC tickets explaining the different issues, so we can close them separately.

Note to Maxim, Ladislav and the Saphiron guys: The multi-script thing would be useful for slim, `include` and encapping. Just saying. Maybe you should push for this.

BrianHawley:
You can apply it now. No need to hold off. That was the last fix.

earl:
No regressions in the rebol-test testsuite. But unfortunately, there are hardly any module-related tests in that testsuite. I didn't try anything else but running the most basic scripts and the rebol-test testsuite.

BrianHawley:
I'll start composing more module tests for the test suite.

carls:
Do we have a bunch of module tests from earlier? I do not remember, but I figure that we must have some tests from earlier design work?

BrianHawley:
We had a module test suite, I think in the old AltME world we were using at the time we designed the current module system a couple years ago. It is likely that most of the tests could be adapted to the current testing model. The main trick is that the module testing suite created and referred to files, which probably won't be OK in the current testing infrastructure.

It should be OK to convert the tests to strings, since loading from strings, binaries, files and URLs is mostly the same, except for decoded values and extensions. The tests of everything other than decoded files and extensions should be adaptable to the new test system.

I'll also have to review the tests to make sure that they are still correct. We'll need more tests for coverage of features that are more recently functional, like exporting from modules.

BrianHawley:
I haven't found the old tests yet. I checked the files of all of the AltME worlds I've been in in the last 5 years (that I have records of) and it wasn't in any of them. I'll start looking through old computers.

Note that this particular pull can be accepted now though.

earl:
Note that a few tests are still embedded [within sys-load.r itself](https://github.com/BrianHawley/r3-bh/blob/b402732/src/mezz/sys-load.r#L701-L765).

BrianHawley:
This PR is blocking further work on the module/script system, especially the experiments we need to run to determine the high-level API for this. We should get it in sooner rather than later.

BrianHawley:
New commit was just a rebase, made because I'm no longer going to wait for this to be merged to continue working on the module system. This will just be included in all future module system improvements.

PR #41, opened 2012-12-18
Fix CC#1865: resolve/extend/only crash

earl:
For RESOLVE/extend, the source context was extended instead of the target context. This mistake can invalidate assumptions and lead to invalid memory accesses, and, in turn, to crashes.

Tested as working and not introducing regressions on Linux 4.4 and Win32 3.1 (using Wine).

CureCode Issue: http://issue.cc/r3/1865


carls:
Excellent.

PR #46, opened 2012-12-18
Fixed #1875 - random/only bug

DanDLee:
The randomly calculated pick position was being added to the block index before being passed to pick.  This caused a pick past the end of the block.

Changed to only pass the pick position.

### DIFF
--- a/src/core/t-block.c
+++ b/src/core/t-block.c
@@ -833,9 +833,9 @@ zero_blk:
 		if (D_REF(2)) Trap0(RE_BAD_REFINES); // seed
 		if (D_REF(4)) { // /only
 			if (index >= tail) goto is_none;
-			index += (REBCNT)Random_Int(D_REF(3)) % (tail - index);  // /secure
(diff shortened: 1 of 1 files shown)


ladislav:
I prefer to use the <tt>Random_Range</tt> function.

DanDLee:
I took a look at the <i>Random_Range</i> function.  It calls the same lower-level <i>Random_Int</i> function. However, it adds bounds checking and sign flipping... stuff that probably isn't necessary when called from a datatype that's already been validated.

ladislav:
<tt>Random_Range</tt> adds other measures, not just those you mentioned. Remainder can have _nonuniform_ distribution in some cases, which is handled by <tt>Random_Range</tt>, but not by the above implementation.

DanDLee:
OK, let's add another ticket to CureCode to request that improvement. Right now, I'm just looking for quick and easy bugs to fix so that I can get familiar with the source code. 

Thanks for your suggestion, Ladislav.

PR #55, opened 2012-12-30
add SET-ENV, tweaks to host api for environment string handling

hostilefork:
This might seem a bit of a non-sequitur.  Like _"why would Fork care about `SET-ENV`?"_

I don't.  It got in my way doing a larger const-correctness-on-string-literals sweep.  Andreas just advised I break it out separately for discussion.

While my const-correcting was mostly just a const here, a const there, no big deal (besides touching 40+ files)...I actually had to rewrite the environment functions a bit to finish it.  But set environment hadn't been used yet, so I couldn't exactly test my changes without fleshing things out.  And as I delved, I discovered _inconvenient truths_ about what a disaster the POSIX environment APIs have been:

http://stackoverflow.com/a/5876818/211160

http://julipedia.meroh.net/2004/10/portability-unsetenvfoo-vs-putenvfoo.html 

This is one of those (not-as-few-as-people-think) cases where Win32 is "less wrong".  So rather than copy POSIX I switched to copying Win32.

Is it "the" answer?  Should the rest of the host API use Win32 API conventions for strings and buffers, or something new?

Err...my impression is that a unified philosophy about the host API has not been hammered out.  So figuring out who-owns-what-buffer and who-should-free-what is something that needs formalism and documentation.  It is tough to abstract abstractions intelligently, and my impression is that the initial open-sourcing of "host-kit" was an attempt to start precisely that discussion.  

It's also my impression that people weren't interested in discussing "host-kit" without an open source story for the Rebol core.  Nowadays I think incentive for a real review might be there, and I'm game.

But this is not meant to be that review or any kind of consistency check.  This is simply meant to unblock the const correctness checkin, without trying to slipstream it in with more uniform-looking changes to 40+ files.  And it adds `SET-ENV`.

Tested on Debian Wheezy 64-bit and MinGW on WinXP.  _Note that POSIX environment keys are case-sensitive and Windows is not._

### DIFF
--- a/src/boot/natives.r
+++ b/src/boot/natives.r
@@ -872,6 +872,12 @@ get-env: native [
 	var [any-string! any-word!]
 ]
 
+set-env: native [
+	{Sets the value of an operating system environment variable (for current process).} 
+	var [any-string! any-word!] "Variable to set" 
+	value [string!  none!] "Value to set, or NONE to unset it"
+]
+
 list-env: native [
 	{Returns a map of OS environment variables (for current process).}
 ]
--- a/src/core/n-io.c
+++ b/src/core/n-io.c
@@ -757,18 +757,67 @@ chk_neg:
 ***********************************************************************/
 {
 	REBCHR *cmd;
-	REBCHR *result;
+	REBINT lenplus;
+	REBCHR *buf;
 	REBVAL *arg = D_ARG(1);
 
 	Check_Security(SYM_ENVR, POL_READ, arg);
 
 	if (ANY_WORD(arg)) Set_String(arg, Copy_Form_Value(arg, 0));
 	cmd = Val_Str_To_OS(arg);
-	result = OS_GET_ENV(cmd, 0);
-	if (result == 0) return R_NONE;
-	Set_String(D_RET, Copy_OS_Str(result, LEN_STR(result)));
+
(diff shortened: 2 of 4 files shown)


carls:
Hey... some interesting code to consider. Let me take a look. (BTW, I thought R3 had a set-env native... hmm, maybe I missed a merge a few years back. Well, glad to see it back.)

Also, would be good if you can some folks to test it on BSD, OSX, older Linux to make sure it's good and portable. I don't really insist on Solaris, Irix, or HPUX anymore... but hey, if anyone is around to try it, might as well.

On const, you're going to find I don't really care for it. My experience has been most of the time it's just a pain, makes code more complex, and I'm not sure there's that much of a win. But hey, as a compiler guy I know the theory of it, and I can be convinced with some good examples of how much speed boost I'd get, right? If we decide to do it... the code changes are going to get ugly for a bit. Like you say, lots of files.

BTW, the const discussion should be made a separate thread from set-env.

earl:
Tested on OSX: builds ok, basic get-env/set-env works fine, no regressions in the testsuite.

Tested on the oldest Linux I have around (Debian "etch", Kernel 2.6.26, glibc 2.3): builds ok, basic get-env/set-env works fine, no regressions in the testsuite.

carls:
2.6.26 is good. Thanks. Merging.

PR #56, opened 2012-12-31
Baby steps toward a pre-commit script, just checks for "pure" tabbed indentation

hostilefork:
Developing cross platform in Rebol (possibly using OSes and editors with which one is unfamiliar, and copying and pasting between virtual machines) introduces lots of possibilities for improper tabbing etc.  Had it happen to me twice now.

_Detab me once?  Shame on you.  Detab me twice?  Shame on me._

Given that git actually supports [pre-commit hooks](http://git-scm.com/book/en/Customizing-Git-Git-Hooks), it would be nice to offer one.  Or at least a step someone could run manually to dot their Indents and cross their Tabs.

As the journey of 1,000 miles begins with a single step, I started with the simplest: look for lines that indent with something other than pure tabs.

Since no one is running it yet, the existence of this checker does no harm.  It points out some cases that exist in [14 files of the current codebase](https://gist.github.com/4418893)...the most extreme in generated files.  Obviously it can be crafted to ignore specific files or enforce other policies if there's a good reason not to use tabs in `boot-code.r` etc.

There's tremendous room for improvement and expansion.  But the immediate goal is just to get the ball rolling so that we have one of these around so as to keep discussions focused on code, and not formatting.

### DIFF
--- a/src/tools/make-commit.r
+++ b/src/tools/make-commit.r
@@ -1,0 +1,182 @@
+REBOL [
+	System: "REBOL [R3] Language Interpreter and Run-time Environment"
+	Title: "Pre-commit consistency checks."
+	Rights: {
+		Copyright 2012 REBOL Technologies
+		REBOL is a trademark of REBOL Technologies
+	}
+	License: {
+		Licensed under the Apache License, Version 2.0
+		See: http://www.apache.org/licenses/LICENSE-2.0
+	}
+	Purpose: {
+		Run pre-commit tests on the codebase for any violations of coding standards.
+		This includes cosmetic issues like spaces instead of tabs.
+	}
+	Note: [
+		"This runs relative to ../tools directory."
+	]
+]
+
+
+; While the coding style and checkin rules should be documented at length in a wiki
+; somewhere, the "success" message should evolve to bring to mind whatever mistakes
+; turn out to be the most common that the tests don't catch
+
+success-message: {
+The source files passed what the pre-commit suite currently tests for!  But also be
+sure that:
+
+* A clean build (make clean, make make, make prep, make) succeeds with zero
+  compiler warnings
+
+* You tested your changes on your operating platform to make sure they do what you think
+
+* You have considered the impact of your changes on platforms you have NOT tested them
+  on.  If it seems likely that impact will be there, then make sure either you (or
+  someone else) does those tests on your branch before you send a pull request to
+  the master.
+}
+
+
+; The failure message shouldn't have to say much, because the alerts delivered by the
+; check should speak for themselves...
+
+failure-message: {
+The source files did not pass what the pre-commit suite currently tests for.  Please
+fix these issues before submitting a pull request for your commit!
+
+(Alternatively: if you believe the tests *themselves* are in error, take a look in
+%/src/tools/make-commit.r and present a suggested modification to the tests along
+with your grievance.)
+}
+
+
+; make-make.r uses the information in file-base.r
+; however this needs to check header files and other things not mentioned there
+; so easiest for now is to just enumerate directories looking for [.c .r .h]
+
+; BUT in case people have some miscellaneous directories not related to the
+; current Rebol /src/ files, we only scan the "known" toplevel ones in the repository
+
+toplevel-dirs: [%boot/ %core/ %include/ %mezz/ %os/ %tools/]
+
+
+;
+; Global variables
+;
+
+num-checked-files: 0
+num-problem-files: 0
+
+
+;
+; Worker functions
+;
+
+check-basic-indent: func [lines [block!]] [
+	errors: copy []
+	line-number: 0
+
+	foreach line lines [
+		++ line-number
+
+		if empty? line [continue]
+
+		whitespace: charset reduce [space tab]
+		non-whitespace: complement whitespace
+
+		unless parse line [any tab non-whitespace to end] [
+			append/only errors rejoin [
+				line-number {:} space {Indentation of line not done with pure tabs.}
+			]
+		]
+	]
+
+	return errors
+]
+
+check-embedded-documentation: func [lines [block!]] [
+	errors: copy []
+
+	; just a stub for the moment...this has an indentation convention also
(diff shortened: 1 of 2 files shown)


carls:
Ok, good point, so let's back up a step and talk about tabs in general. Should we do it here or on some msg board, I don't care. Maybe the blog again?

Here's the starting point: indents are 0x9 tab hardcoded with visual display tabspace = 4.  Main reason: easier for dumb editors (that don't do auto-space inserting tabs) and it also produces smaller files, if that matters to anyone but me these days.

Now, that said, years ago Gabriele suggested that .r files use spaces, not tabs. Reason: more consistent formatting of Rebol scripts online. It's a good point, and I somewhat agree, but probably don't abide by it, which is why you probably find various .r files in the source using tabs.

I think for the earlier R3 host releases I would standardize the format, but I don't remember, it was so long ago.

So, tell me what you think... and let's get this behind us (as I'm eager to see some graphics come online again.)

BrianHawley:
I prefer tab indenting. It was annoying switching between Rebol-standard tab indenting and Qtask-standard 4-space indenting when I was working on both projects. Had to write conversion scripts to switch between the two.

meijeru:
The Crimson editor has easy conversion between tabs and spaces, and I suppose many more "programmers' editors"
have such facilities.

hostilefork:
_(Note: Thanks to Andreas I un-entangled this commit from the environment one.)_

As BrianH noted, my test itself had multi-line strings geared for console output that didn't follow the rule.  (Of course, I was purposefully separating out pre-commit script from modifying files into alignment with the rules, so I wasn't really looking too closely at the output...precisely because a discussion like this would probably happen.)

I used to be one of those _"you're crazy if you use spaces instead of tabs"_ people.  Now that more and more code is being typed into the web, where browsers use the tab key to move around from control to control, I'm less sure.  Having tabs starts seeming a little bit like the kind of situation you wind up in when Microsoft Word turns your quote characters from ordinary ones into typographic curly ones...and once they start creeping in they screw everybody up.

By contrast, though, tabs are rather foundational and agreed-upon ASCII characters.  I've been using them and it would be easier to not change.

Guess I don't really have a horse in that race, other than the obvious _"making a decision sooner is better than making it later"_ and _"checking things automatically is better than having the conversation in a pull request"_.  In spirit this isn't `fix-the-indentation`, it's `pre-commit-script`.  :-)

earl:
I prefer space-based indenting, precisely for the reason you quote as given by Gabriele: consistent visual appearance. With the limited display systems we have today, you always have to carry an "intended width" along when using tabs, or live with the displayed results looking awkward or even outright ugly on many systems.

So again: I'd strongly prefer to switch to space-based indentation (with 4 spaces) for all files, including .r, in the R3 sources.

ladislav:
My preference is tab-based indenting

carls:
I've always preferred tab-based indenting because the semantic (or maybe "intention" is a better word) is explicit; it's sort of like using a blockquote in HTML. There's more meaning to it than just so many spaces.

Also, at the ReCode conference, I think it was onetom who pointed out how poor spaces looked in his favorite editor, where he prefers to use proportional fonts for readability.

In my opinion, any system that converts text to web content should do the "proper thing" and convert either by ruler tab spacing or with tab to space conversion.

onetom:
## At the conference

Yup, that was me. You seemed to like my font, right? ;)
It was [Muli](http://www.fontsquirrel.com/fonts/download/muli). I'm using [Ubuntu](http://font.ubuntu.com/download/ubuntu-font-family-0.80.zip) mostly, but "Verdana" is very good too.
I'm happy using proportional fonts for code editing since the Mac version of Sublime Text was release (2+ years ago).

## The predominant usage of proportional fonts

In fact, most people - including programmers - use (read&write) proportional fonts during most of their life, since webpages, emails, comments (like this one itself), SMSes, AltMe ;), etc etc are all rendered with proportional fonts.

While we are doing math by hand in primary school, we didn't switch "fonts" either... As kids we had to practice on squared notebooks - practically in monospace - but quickly we all gave up on the square boundaries... I remember we could even buy french "squared" notebooks, which had a grid of standing rectangles instead of squares...

## Showing tab on the web

Presenting on the web is not a big problem actually. I have this user stylesheet to show tabs as 2 spaces (on GitH*b for example):

```
$ cat ~/Library/Application\ Support/Google/Chrome/Default/User\ StyleSheets/Custom.css
pre {tab-size: 2}
```

## Typing tab on the web

Have a look at https://coderpad.io/
They are modern, online, collaborative code editor and they have a tab size selector in the upper right hand corner... They simply use spaces though, but think about it; what is the drive behind such an option?

They don't have a problem interpreting the TAB key to let it insert a tab character, neither do Google Docs have any problem to make TAB jump to the next left/center/right tab-stop OR within lists increase indentation level...

## "Semantic" aspect or what is the intent?

As Carl pointed out, tab is used to express some meaning. It allows different - user preference specific - visual interpretation (as in how much space does it actually mean).

But the possibilities doesn't end here! Have a look at the idea of [Elastic tabstops](http://nickgravgaard.com/elastictabstops/) No more silly fiddling with spaces to line up code to create tables (which keep on getting misaligned as new rows are entered, plus they don't work with proportional fonts either at all...)

(Character graphics is not compatible with proportional fonts either; that's a drawback which I have no sane workaround for)

## Conclusion

Keep on using tabs.

When elastics tabstops becomes more supported, we will be thankful for this decision and we can ripe the benefits right away.

hostilefork:
Brought this up with @DocKimbel, he has voiced his support for tabs and said: "I've always used tabs and see no good reason to change it."

http://chat.stackoverflow.com/transcript/291?m=11466493#11466493

So with Carl in favor of tabs, both language principals have spoken.  I myself would lean toward a transition to spaces, but this is one of the many "hard" decisions we have to just make.  Consistency across our toolsets and editor settings to make these projects go forward is more important than any imagined benefit the difference might have.

Forward motion is necessary.  Let's standardize Rebol projects on tabs, fix it everywhere, and _never discuss them in a pull request again_.

PR #57, opened 2012-12-31
Fix missing /local in make-boot.r

earl:
Due to the missed /local, the global `data` is overwritten. Currently,
this only results in a weird informational message (compressed code
beeing much larger than the supposedly uncompressed code).

But it also result in the uncompressed original code becoming
unavailable after compression. This is a nuisance, if we want to emit
uncompressed code for whatever reason.


carls:
Ah... good catch on missed local.

PR #63, opened 2013-01-14
Match  "uncased" parse in Check_Bit with !HAS_CASE(parse)

hostilefork:
This fixes [CureCode issue #1938](http://issue.cc/r3/1938)

The definition for [Check_Bit]() is:

```
REBFLG Check_Bit(REBSER *bset, REBCNT c, REBFLG uncased)

Check bit indicated. Returns TRUE if set.
If uncased is TRUE, try to match either upper or lower case.
```

However the call from PARSE when matching against a bitset passes in [HAS_CASE(parse)](https://github.com/rebol/r3/blob/master/src/core/u-parse.c#L216).  The `/case` refinement on parse indicates a desire to _pay attention_ to case, not to ignore it.  This led to the reverse of the desired behavior:

```
>> ab: charset "Ab"
== make bitset! #{00000000000000004000000020}

>> parse "AB" [2 ab]
== false

>> parse/case "AB" [2 ab]
== true
```

Inverting the condition corrects this.  It also fixes issues with parsing binary data...where `/case` is always implied (thus could _never_ be case-insensitive as written).

### DIFF
--- a/src/core/u-parse.c
+++ b/src/core/u-parse.c
@@ -213,7 +213,7 @@ void Print_Parse_Index(REBCNT type, REBVAL *rules, REBSER *series, REBCNT index)
 
 	// Do we match to a char set?
 	case REB_BITSET:
-		flags = Check_Bit(VAL_SERIES(item), GET_ANY_CHAR(series, index), HAS_CASE(parse));
+		flags = Check_Bit(VAL_SERIES(item), GET_ANY_CHAR(series, index), !HAS_CASE(parse));
 		index = flags ? index + 1 : NOT_FOUND;
 		break;
 /*


earl:
Tested against the [core-tests testsuite](/rebolsource/rebol-test): no regressions.

hostilefork:
This pull request is now one week old.  Others are older.  :-(

Integrating this patch is particularly important--because the bug completely blocks the current most active Rebol codebase from using Rebol 3 instead of Rebol 2.

Of similar import is the integration of [pull request 60](https://github.com/rebol/r3/pull/60), because it fixes (among other things) a break to `make make` that causes Rebol to be unable to be used for bootstrap.

Hopefully Carl is okay.  (!)  Assuming he is, the next most important issue is how we can get merges of fixes in a timely manner, so that the project can move forward...!!!  Carl had suggested a deputization of someone(s) to be responsible for these integrations, and it seems to me that it might be time to make those choices.  _(That was not me volunteering.  :-P)_

carls:
Thanks for the fix. I was in the middle of work work. Requires a lot of attention at certain times.

PR #66, opened 2013-01-19
Fix crash when reading dns:// without a hostname

earl:
Previously, `read dns://` crashed the interpreter. With this fix, a dns:// spec without a hostname (or IPv4 address) is considered invalid for READ access:

```
>> read dns://
** Access error: invalid spec or options: dns://
```

Fixes the crash reported in [CureCode issue #1935](http://issue.cc/r3/1935).


Oldes:
This fixe introduced a regression that it's not possible to resolve a domain name from IP anymore!

PR #69, opened 2013-01-21
bug#1939

ladislav:
accuracy of mold and mold/all for negative zero
tested using core-tests


carls:
Thanks for -0.0 fix. On Parse, hopefully it will pass regression tests.

PR #72, opened 2013-01-24
String literals as const char * instead of char *

hostilefork:
This isn't a const-for-the-sake-of-constness push.  This is just living with the reality that modern compilers do not like:

```
char *str = "foo";
```

The reason is that string literals (in both C and C++) cannot be modified.  Therefore, this statement would yield undefined behavior:

```
str[0] = 'g';
```

This is why C++ compilers (and C compilers with warning levels raised up) will gripe unless you write such declarations as:

```
const char *str = "foo";
```

All this commit does is start from doing this to all the string literals in the codebase, and slogs through the transitive repercussions.  Obviously that means touching a lot of prototypes.

No bugs found--at least I don't think any were.  But it protects against them creeping in as others modify the code w/o understanding where data came from.  Also, a couple of surprising things popped up that are probably good to know about.  For instance, if you call the Unicode version of CreateProcess it [modifies the string you pass in](https://github.com/rebol/rebol/pull/72/files#L38R596).  :-/  Who knows what they were smoking.

The other "please take a closer look at this" codepoint would be the allocation here:

https://github.com/rebol/rebol/pull/72/files#L36R139

With these changes (and a few more I'll queue up later), Rebol can be built with C++ compilers with a warning count that I'll call....ermmm... _"manageable"_.


earl:
If we want const str literals as a general goal (and policy), we should also have it checked by the compiler, otherwise the cleanliness established by this pull request will quickly degenerate again. For GCC, that means enabling [`-Wwrite-strings`](http://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Warning-Options.html#index-Wwrite_002dstrings-405).

hostilefork:
Superseded entirely by [Coherence 1.5](https://github.com/metaeducation/ren-c/pull/12) which is way, way further ahead.  Closing.

PR #73, opened 2013-01-24
Use David M. Gay's dtoa for molding decimals

ladislav:
Currently, R3 uses platform-specific code for molding decimal numbers.
This increases the effort necessary to port to new platforms like
android-arm or linux-arm

Porting R3 to android-arm we found none of the system-specific methods
currently used by R3 were available. Instead of adding another
system-specific method specifically for android-arm, we decided to
replace the system-specific methods using the dtoa function
implementation by David M. Gay.

David M. Gay's original code is available at:
http://www.netlib.org/fp/dtoa.c

This implementation brings the following advantages:
- MOLD and MOLD/ALL results will be accurate (solving cc#897).
- MOLD results will be consistent across wide range of platforms
  (solving cc#1633).
- MOLD results will be optimal: yielding the shortest string that can be
  used to reproduce the number (i.e. a string that is accurate enough).
- Makes porting to new platforms easier.
- Battle-tested solution, as this is the way how the problem was solved
  in e.g. Java, Javascript, Python, Firefox, Chrome.

The commit makes the following changes:
- David M. Gay's implementation is added as src/core/f-dtoa.c. The
  source is unmodified except for an #include of Rebol-specific
  configuration and a reference to where the source was downloaded from.
- An attribution notice is added to NOTICE.
- Configuration for f-dtoa.c is contained in src/include/reb-dtoa.h.
- The additional source is listed in src/tools/file-base.r.
- make/makefile is re-generated to accommodate the additional source.
- The HAS_ECVT, HAS_LOG10L, HAS_LONG_DOUBLE Rebol configuration defines
  are removed as they are no longer necessary.
- HAS_LL_CONSTS was previously implied by HAS_LONG_DOUBLE; it is now
  defined explicitly for the respective platforms.
- Functions for decimal conversion are modified to use dtoa.

This resolves CureCode issues 897, 1633, 1942.


earl:
Tested successfully on Linux-ARM, Linux-x86, OSX-x86, and Win32-x86.

ladislav:
Tested successfully on Android-ARM

ladislav:

onetom:
Guys, saying "Tested successfully" is not very reassuring, especially when there is no test code in the commit.
Since the corresponding test suite is somewhere else, it would be nice to link in the relevant commit, so others could review, replicate and learn from the the tests later.

ladislav:
onetom, could you be a bit more specific? It is true that the core-tests suite is in a different GitHub repository, but I don't know what your suggestion/idea is.

earl:
@onetom "Tested successfully" means that we ran the [core-tests testsuite](https://github.com/rebolsource/rebol-test) without any regressions.

In this particular case, there are probably tens to hundreds of related tests in the testsuite, many of them written long before this fix was created. So linking to  a single relevant commit will prove hard to impossible in this case.

However, I agree that it would be useful to integrate the tests more closely. Maybe by adding links to a pull request to rebol-test commits added in support of the specific pull request (quite a bit of manual work, I fear). Or at least by referencing the current head revision of rebol-test used for regression testing, when posting "tested successfully" messages such as the above.

ladislav:
Aha, thanks for clarification, earl. I agree, it is hard to impossible since the tests are usually created long before the pull request is, thus, there actually is no link between pull requests and tests.

PR #76, opened 2013-02-02
Trim that bug! (src/core/s-trim.c)

CCX:
2013.02.02 by CCX

```
>> trim "  x ^/"
== "x x^/"
```

Fixed that trim bug, found in `trim_head_tail()`. Changed some var
names to make the algorithm easier to understand. `head` and `tail`
are now trimmed first because its simpler this way.

First time I ever use github...

### DIFF
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -239,8 +239,7 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 			}
 			else if (uc == LF) {
 				outside = TRUE;
-				if (left) out = left;
-				left = 0;
+				if (left) out = left, left = 0;
 			}
 			else {
 				outside = FALSE;
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -204,7 +204,7 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 ***********************************************************************/
 {
 	REBCNT out = index;
-	REBOOL alf = FALSE; // append line feed
+	REBOOL append_line_feed = FALSE;
 	REBUNI uc;
 
 	// Skip head lines if required:
@@ -219,7 +219,7 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 	if (t || !h) {
 		for (; index < tail; --tail) {
 			uc = GET_ANY_CHAR(ser, tail -1);
-			if (uc == LF) alf = TRUE;
+			if (uc == LF) append_line_feed = TRUE;
 			if (!IS_WHITE(uc)) break;
 		}
 	}
@@ -259,7 +259,7 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 	}
 
 	// Append line feed if necessary
-	if (alf && !t) {
+	if (append_line_feed && !t) {
 		SET_ANY_CHAR(ser, out, LF);
 		++out;
 	}
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -230,9 +230,9 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 		REBCNT left = 0; // index of leftmost space (in output)
 
 		for (; index < tail; ++index) {
-			
+
 			uc = GET_ANY_CHAR(ser, index);
-			
+
 			if (IS_SPACE(uc)) {
 				if (outside) continue;
 				if (!left) left = out;
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -209,7 +209,7 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 
 	// Skip head lines if required:
 	if (h || !t) {
-		for (; index < tail; ++index) {
+		for (; index < tail; index++) {
 			uc = GET_ANY_CHAR(ser, index);
 			if (!IS_WHITE(uc)) break;
 		}
@@ -247,14 +247,14 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 			}
 			
 			SET_ANY_CHAR(ser, out, uc);
-			++out;
+			out++;
 		}
 	}
 	else {
-		for (; index < tail; ++index) {
+		for (; index < tail; index++) {
 			uc = GET_ANY_CHAR(ser, index);
 			SET_ANY_CHAR(ser, out, uc);
-			++out;
+			out++;
 		}
 	}
 
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -196,14 +196,10 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 
 /***********************************************************************
 **
-*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG const H, REBFLG const T)
+*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG h, REBFLG t)
 /*
 **		Trim from head and tail of each line, trim any leading or
 **		trailing lines as well, leaving one at the end if present
-**
-**	2012.02.02 CCX: Fixed the trim bug. Changed some var names and
-**		the function declaration to make easier to understand the
-**		algorithm.
 **
 ***********************************************************************/
 {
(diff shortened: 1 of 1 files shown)


src/core/s-trim.c
```diff
@@ -196,76 +196,79 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 
 /***********************************************************************
 **
-*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG h, REBFLG t)
+*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG const H, REBFLG const T)
```
hostilefork: Welcome to GitHub!  :-)

All things being equal, it would be nice if const was the default and mutable was explicit.  But this is a case where I think const just makes the code longer without a benefit in proportion to the extra characters.

I mention this because Carl is a `const` skeptic to begin with, and I'm trying to get a pull request in that's based on what I consider to be an important case w.r.t. undefined behavior:

https://github.com/rebol/r3/pull/72

So in the spirit of maintaining a sense of perspective regarding the usage of const, I'd encourage you to remove these!

src/core/s-trim.c
```diff
@@ -196,76 +196,79 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
...
+			++out;
 		}
 	}
 	else {
-		for (; index < tail; index++) {
+		for (; index < tail; ++index) {
 			uc = GET_ANY_CHAR(ser, index);
 			SET_ANY_CHAR(ser, out, uc);
-			out++;
+			++out;
```
hostilefork: Moving the location of the increment on a line where it does not make a difference is what we would call "code churn".  :-)  While we all have our preferences, change for the sake of change makes the commit harder to review.  Format is secondary to function, so if you see any cases where Git is adding and removing blank lines (for instance) try and discard that before the pull request.

P.S. Please feel free to join the [Rebol/Red StackOverflow chat room](http://chat.stackoverflow.com/rooms/291/rebol-and-red)!

src/core/s-trim.c
```diff
@@ -196,76 +196,79 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
...
 **
-*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG h, REBFLG t)
+*/	static void trim_head_tail(REBSER *ser, REBCNT index, REBCNT tail, REBFLG const H, REBFLG const T)
 /*
 **		Trim from head and tail of each line, trim any leading or
 **		trailing lines as well, leaving one at the end if present
 **
+**	2012.02.02 CCX: Fixed the trim bug. Changed some var names and
+**		the function declaration to make easier to understand the
+**		algorithm.
```
hostilefork: Oh...and you don't need to add in-source change logs. The version control system already takes care of this!  Yay GitHub!

_(Check out the "blame" feature if you have not... finds the origin commit of every line in a file!  :-P)_

CCX:
EDITED — Some comments are now closer to the source code.

I have not found (yet) how to edit my pull request. Can it be edited, or should
a new one be submitted?

•CCX•

BrianHawley:
CCX, you can add new commits to the same branch that change stuff you want changed, and as long as the pull request hasn't been accepted yet it will be added to the request.

src/core/s-trim.c
CCX: I think `x++` possibly adds some unnecessary overhead to `++x`, so I don't consider it a matter of taste.

CCX:
Thanks BrianHawley, that was really helpful.

•CCX•

src/core/s-trim.c
CCX: Hi. Thanks for your comment (cool name by the way.)
Well, yes, I guess `consts` are kind of useless in this case.

src/core/s-trim.c
hostilefork: [Is there a performance difference between i++ and ++i if the resulting value is not used? (StackOverflow)](http://stackoverflow.com/questions/24886/is-there-a-performance-difference-between-i-and-i-in-c)

"Executive summary: No."  :-)

src/core/s-trim.c
CCX: Thanks for the tip.
•CCX•

earl:
First, thanks for your contribution! While we may seem picky about process, we really appreciate your help.

We aim to keep each change as small as possible in order to simplify code review. The pre-/post-increment change we were discussing inline is irrelevant to the bug you are trying to fix, and thereby goes beyond the minimal change necessary; this we call "code churn". So for bug fixes, please try to keep the changes as minimal as possible.

earl:
Retroactively added [CureCode ticket #1948](http://issue.cc/r3/1948) for this bug.

CCX:
Yes you all seem very picky... you got that part very clear!

(And Its starting to make me rethink about collaborating at all to be honest.)

I hand coded the function almost completely from zero to spot the bug, that would explain minor _irrelevant_ differences in the source code. At the end, the source with the fix was very similar to the original, so I edited to make it look as close as the original as possible. Knowing how maniac programmers can be about coding styles, I was very careful... but it seems it was not enough.

I believe the source code I willingly submitted fixes the bug in the `trim_head_tail()` function. I already dedicated enough time fixing the bug and defending my source code. Whoever reviews my pull request is free to accept, modify, or reject it, as his time permits.

•CCX•

(This comment was a Line Note.)

hostilefork:
Don't be stressed, CCX!  :-)  Nice to have more people getting involved!

Collaborating and getting feedback can be a lot of fun, and is how we all learn things.  Perhaps you haven't run across a work environment where these sorts of rules are in place, but they aren't made up on a whim...they come from the school of hard knocks.  Especially when people have to deal with big codebases where seemingly minor code reorganizations can have unintended consequences.

@earl pointed out that the question of whether `++x` is better than `x++` is independent of the bug.  I'm a C++ person so in that spirit I would support using ++x, as it can make a small difference there in theory.  But Rebol is a C codebase and it simply isn't going to be seen as a priority on this project, where all optimizers will generate equivalent output code.

But if it were worth doing then it would be worth doing in one large commit, to _all_ the code.  It would be done with a thorough search-and-replace, and identified in the version history as "The x++ to ++x mass change."  This would make it easy to review, and people wouldn't have to shift gears as they scrolled through the diffs.  If someone didn't care about the issue they could skip reviewing it without worrying they were missing something they _did_ care about.  And if it turned out to be a bad idea, all of them could be undone en-masse with version control automation (without accidentally losing your trim fix in the process!!)

These aren't huge deals for a very small patch and might seem like making a big deal out of nothing, but it's really just to set a bar for the process that we all agree to adhere to...for changes big and small.  If you want to chat about the general philosophy and Rebol in general, please [join us in the chat room I mentioned](http://chat.stackoverflow.com/rooms/291/rebol-and-red), and we can get to know who you are...!

earl:
@CCX, I submitted a slightly reduced version of this pull request without the post-increment to pre-increment changes as #111. Proper attribution of the fix to you is of course left in place. I suggest to close this pull request, and use #111 instead.

CCX:
I still cannot find how to avoid some `+ empty-lines` and `- empty-lines`.
Now it looks clean enough (at least for me).

earl:
Again, thanks for your work on this. Please don't be too annoyed by the requests for cosmetic improvements. If you don't want to be bothered at all with this, that's fine as well; we can also take it from here.

If you're still willing to do the clean-ups yourself:

The "empty line" changes are because you add or remove trailing whitespace on a line. In this particular case, you are _adding_ trailing whitespace on 7 lines; please undo that. Plain `git diff` (with colour enabled: `git config --global color.ui auto`) will highlight trailing whitespace, if your editor doesn't have a facility to do so. Github unfortunately seems somewhat random at showing this (at least for me): [line 235](https://github.com/rebol/r3/pull/76/files#L0R235) is one case where Github shows this as well (for me).

There's one post-increment to pre-increment change left in [line 232](https://github.com/rebol/r3/pull/76/files#L0R232); please change that back as well.

For intra-file consistency, please also rewrite the other pre-increment uses in loops to post-increment: lines [220](https://github.com/rebol/r3/pull/76/files#L0R220), [264](https://github.com/rebol/r3/pull/76/files#L0R264).

Finally, with this done, please [squash all those commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) into a single one.

All of this may seem like just uselessly wasting time. However, a consistent and clean code-base is an extremely valuable asset and greatly eases working with and collaboration over this code-base. My hope is that we can all work together to at least maintain the current level of consistency and cleanliness.

earl:
Thanks a lot, @CCX, for cleaning up the remaining cosmetic issues. [Squashing](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) would still be nice, but as far as I'm concerned, this pull request is good to merge.

PR #77, opened 2013-02-05
Fix recursive make calls

earl:
Instead of hardcoding the name of the make command, use the predefined `$(MAKE)` variable for recursive make invocations.

This fixes the build for environments where make is called something else (like `gmake`, or `mingw32-make`). `$(MAKE)` is standardised in [POSIX make](http://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html#tag_04_84_13_08) and present in at least [GNU make](http://www.gnu.org/software/make/manual/make.html#Recursion) and [BSD make](http://www.freebsd.org/cgi/man.cgi?query=make&manpath=FreeBSD+9.1-RELEASE#VARIABLE_ASSIGNMENTS).

Problem most recently brought to attention by Adrian Sampaleanu.


ladislav:

PR #78, opened 2013-02-06
Fix crash when accessing an empty but hashed map

earl:
A `map!` internally switches to a hashed implementation if the map holds more than a given number of items. If a map becomes empty again after having switched over to hashing, lookups on the empty map lead to a division by zero (in modular arithmetic), crashing the interpreter.

This commit fixes the division by zero crash by side-stepping modular arithmetic for the special case of an empty map.

The problem was originally identified by user HostileFork on Stack Overflow chat.

Fixes [CureCode issue #1930](http://issue.cc/r3/1929).

### DIFF
--- a/src/core/t-map.c
+++ b/src/core/t-map.c
@@ -118,9 +118,9 @@
 	if (!hash) Trap_Type(key);
 
 	// Determine skip and first index:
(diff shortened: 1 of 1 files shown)


ladislav:

PR #80, opened 2013-02-10
Make APPLY secure against RETURN/redo (#1949)

BrianHawley:
See http://issue.cc/r3/1949 for details.

### DIFF
--- a/src/core/n-control.c
+++ b/src/core/n-control.c
@@ -293,6 +293,7 @@ enum {
 ***********************************************************************/
 {
 	Apply_Block(D_ARG(1), D_ARG(2), !D_REF(3)); // stack volatile
+	VAL_CLR_OPT(DS_TOP, OPTS_REVAL); // secure against return/redo
 	return R_TOS;
 }
 


BrianHawley:
Note: We would still want to include this fix if we got rid of RETURN's /redo option. It would still be a problem for values like this that are created in native code. These values are not an error, they are part of the normal process of DO function. The only problem is that people use the APPLY native to more safely call functions without getting hacked. This is why I put the fix in APPLY and not in Apply_Block, because it is only when calling APPLY that this is not a normal value.

BrianHawley:
OK, so I summarized a little in the commit message. It is security against unwind values of the return type, which have referenced function values (except ops), and have the `OPTS_REVAL` flag set. These values happen normally in R3 code, particularly when you use the `do` function, and are not themselves erroneous. Other native functions would not be in error to create such values.

For convenience in demonstrating the creation of these values I have used `return/redo` in the code examples. There is nothing wrong with `return/redo` (well, see http://issue.cc/r3/2121 but that doesn't matter here), it just made a convenient way to show the creation of such values. I could have just as easily used the `do` function with no refinements to get the same effect.

Here is the example code rewritten to use `do`:

``` rebol
; Current behavior
>> apply :do [:add] 1 1
== 2

; Preferred behavior
>> same? :add apply :do [:add]
== true
```

In this case, for convenience, I have referenced `do` explicitly but the same would work in cases where someone is using `apply` to limit the access of a function it is calling, and someone snuck a `do` reference in there. If you would like to remove the `do` function as well instead of this simple fix, good luck with that.

Without this fix it is quite trivial to use `do` to enable a get-word hack to get references to all code in the block after the `apply` call, without affecting evaluation of the function. This basically makes `apply` security _completely useless_, regardless of whether `return` has a `/redo` option or not.

BrianHawley:
We want to block the _automatic part_ of the redo completely. The returned value can _and usually would_ take a different argument spec, so it would be rare that the same `apply` values block would work with both.

If you are actually expecting a redo function to be returned, you are most likely going to want to `apply` it separately to give it a different set of arguments. Redos reevaluate their arguments too, and with different specs the expected values block could be quite different even if they take the same number of arguments (which they don't have to). Redos can take arguments with different evaluation behavior, and certainly the typespecs could be different. That is the whole advantage of redos, otherwise there would be no point to them.

More often you will just want to block the automatic redo altogether, because redos are not as useful when you're picky or paranoid enough to be using `apply` in the first place. They're only useful when you know what they're going to do, same as with `do` function.

BrianHawley:
You need to stop thinking of redos as just a security hole. Like function values though, they can be a security problem in certain rare circumstances, and it is important to note that these are the _same_ circumstances where function values can be a problem. As such, it makes sense to make sure that the methods that we use to secure function values can be used to secure redos as well.

When calling a function value so it can do its job - but not more than that - is what you need to do, you use `apply` to call the function. That makes the function value useful to you even when it might potentially be trying to hack your code. It makes sense to take the same approach with redos, making it possible to take advantage of the rare potentially malicious but useful redos. However, the full redo process corresponds to _two_ calls to `apply` with _different_ blocks of values. And if you're _not expecting_ a redo, then _you don't want the redo to happen at all_ because it is almost certainly malicious.

That is the balance that makes redos still useful in paranoid situations, which aside from wrapper functions is the main purpose for which `apply` is used.

BrianHawley:
Sorry that you didn't understand (or are being silly, I can't tell).

I mean that one of the main reasons to use `apply` in the first place is if you're already picky or paranoid about the situation (usually for good reasons), because `apply` is supposed to _solve_ the problems that you are picky or paranoid about, by letting you have better control of the function parameter evaluation process. I certainly didn't say that `apply` _makes_ you picky or paranoid, I said exactly the opposite of that.

The other main reason to use `apply` is to pass refinements along to other functions in function wrappers like `repend`.

BrianHawley:
The other reason you want to block automatic redos _if you don't expect them_ is that they can be written to never stop, if they are created by functions which don't have the strict constraints of DO _any-function_ (like `return/redo`).

Each redo can return another redo. If that is what you want, sure, but otherwise it's a denial of service attack, and not even wrapping them in parentheses can stop that. If we don't have a way to stop them, `apply` or even parens won't work as security measures. Not as big of an issue as get-word attacks, but still worth stopping even though having forever loops in them would do the same thing.

Nonetheless, it's the different specs issue that makes `apply` redo with the same value block a bad idea.

hostilefork:
@Ladislav please see this SO question and if you really think this is a bad idea, say why.  But I really like it as a user feature.

[Why does return/redo evaluate result functions in the calling context, but block results are not evaluated?](http://stackoverflow.com/questions/14760916/why-does-return-redo-evaluate-result-functions-in-the-calling-context-but-block)

Brian has made an argument (I think?) for why maybe it should throw an error on non-functions, just to make sure you know what you're getting.

earl:
Regarding apply+redo behaviour, I would prefer the following behaviour as well:

```
>> print apply :do [:add 1 1] 2 2
2
```

(I know full well that this is in muddy waters with the variadic behaviour of DO. But I think I'd prefer above behaviour regardless.)

However, as a first step to get there, I consider `apply` blocking redos completely (as is done in the fix under discussion)  a definitive improvement over the status quo.

BrianHawley:
Fork, someone might have made an argument that `return/redo` should trigger an error for non-functions, but not me. I'm just fine with the `/redo` option being ignored completely for non-functions. We have more functions in R3 where we ignore inapplicable options instead of triggering errors because this approach makes it easier to write more flexible code.

If it triggers an error, and you might want to return either function types or other types, then you have to wrap the `return` calls in slow conditional code, with one code branch calling `return/redo` and the other calling just `return`. If it just ignores the option then you don't have to write the conditional code, you just write `return/redo` without the conditional code, which in the common case of just not needing the option is much faster.

BrianHawley:
Andreas, Ladislav, you seem to be interested in having `apply` act more like `do`, which pretends that a redo is another part of the original function call, when it is really a _second_ function call to a _different_ function with a _different_ set of arguments. That pretense is good for `do`, but it's a bad idea for `apply`. It is unfortunate that for simplicity I've chosen a harmless set of function arguments to illustrate a security fix that is intended to guard against harmful functions, because it makes it easier to fool yourself into thinking that redos are OK all of the time.

One option that we could do, which would be more flexible, would be to reimplement the redo behavior in `Apply_Block` itself _only when the `do` function is the one being applied_. Then you can have `Apply_Block` deal with as many redos as come up, because `do` is a function that only does redos when it is passed functions, not arbitrarily. With a single exception for the `do` function, even if implemented repetitively for calls to `do do function`, `Apply_Block` would still be safe.

If, on the other hand, any redos escape from `Apply_Block` to the `apply` return value, this is definitely a security problem that can give access to code _outside of the values block_. So at the very least _this_ fix needs to be implemented regardless of any `Apply_Block` enhancement.

ladislav:
This fix improves the status quo, but it does not look as a definitive solution.

ladislav:

BrianHawley:
Deleted or edited comments that were replying to ones that Ladislav deleted later.

Ladislav: "This fix improves the status quo, but it does not look as a definitive solution."

Since that comment was made, we determined that this fix is one part of the only solution we can safely get away with. We determined in other experiments that optional function arguments make combined-block or even the multi-block `apply` infeasible and a bad idea. This fix can and should be merged ASAP.

For the second part, we need to remove the `/redo` option from `return`. See http://issue.cc/r3/2121 and #198  for details.

BrianHawley:
Needs a new PR, for a new ticket.

PR #81, opened 2013-02-11
Fix SET object/block block assuming /any (#1763)

BrianHawley:
SET without /any wasn't triggering the error it should have when assigning a block of values that include unsets to a block or an object. This tests for that condition and triggers the error before even a single assignment happens, all or nothing.

See http://issue.cc/r3/1763 for the ticket.

Also added documentary comment about new unwind policy. The unwind screening required to fix http://issue.cc/r3/1515 must be performed elsewhere, in Set_Var, Do_Next and other related functions, for it to work properly. Not practical to put the fix here.

### DIFF
--- a/src/core/n-data.c
+++ b/src/core/n-data.c
@@ -507,12 +507,14 @@ static int Check_Char_Range(REBVAL *val, REBINT limit)
 **
 ***********************************************************************/
 {
-	REBVAL *word = D_ARG(1);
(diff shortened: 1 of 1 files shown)


earl:
Builds and tests fine, no test regressions. Reviewed the fix: +1

As this is a long-standing "silent" bug, I'm happy to see it finally fixed.

ladislav:

PR #82, opened 2013-02-15
Make DO binary treat the binary as source

BrianHawley:
LOAD and IMPORT treat binary data as UTF-8 source. So should DO.

Also, fixes an outdated comment.

See http://issue.cc/r3/1952 for the rationale for this fix.

### DIFF
--- a/src/core/n-control.c
+++ b/src/core/n-control.c
@@ -548,10 +548,11 @@ got_err:
 		if (IS_THROW(value)) return R_ARG1;
 		Throw_Error(VAL_ERR_OBJECT(value));
 
+	case REB_BINARY:
 	case REB_STRING:
 	case REB_URL:
 	case REB_FILE:
-		// DO native and system/intrinsic/do must use same arg list:
+		// DO native and sys/do* must use same arg list:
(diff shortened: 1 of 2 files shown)


hostilefork:
While some might argue that having DO of a value type that doesn't obviously represent user-readable source code, BrianH said this in chat:

> **do to string! some-bin** has to convert the string back to UTF-8 binary before it can be parsed, because Rebol's source parser can only handle UTF-8 binary. We don't have a string parser for Rebol source. So, this could be a major speedup for a wide variety of situations.

In light of that, this seems like it would be an uncontroversial change.  If you have a UTF-8 binary representing source that you got from somewhere, why not be able to DO it directly, as it would have to be turned into a string and then back into UTF-8 bytes otherwise?

Is there any reason it's not merged (besides no one having commented on it?)

carls:
A few years back, I was picky about the meaning of DO on BINARY!, thinking there might be something else you want done, like playing a video or audio binary, etc.  Since file IO defaults now to binary, it makes sense not to require the conversions just to DO a source binary. So, I'll accept it. We can add a PLAY someday... when that's the kind of DO we mean.

PR #87, opened 2013-02-18
bug#1955

ladislav:
Hostile Fork, Brian Hawley and I are proposing to use the notion of "construction syntax", "construction" or "construction spec" for #[...] like syntax elements.

Reasons:
- constructor (MAKE or something) is actually the part missing in construction specs
- the "serialized syntax" is not appropriate either since it is just a redundant and pleonastic equivalent of "syntax"


ladislav:

PR #88, opened 2013-02-19
bug#1956

ladislav:
Corrects the (previously incorrect) help strings of EXP and LOG-E

### DIFF
--- a/src/boot/natives.r
+++ b/src/boot/natives.r
@@ -788,7 +788,7 @@ arctangent: native [
 ]
 
 exp: native [
-	{Raises E (natural number) to the power specified.}
+	{Raises E (the base of natural logarithm) to the power specified}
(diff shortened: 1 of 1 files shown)


angerangel:
I'm not English, but original descriptions seem correct to me.

BrianHawley:
It's a math terminology precision thing. We had a big discussion on it in AltME. The new docstrings are more precise, without bringing up terms like "antilogarithm" or names like "Euler" which someone who actually _needs_ the docstrings would be not likely to know. It's an improvement.

ladislav:
"original descriptions seem correct to me" - the original description was mathematically incorrect. E is not a natural number, you can look up both e and "natural number" in Wikipedia in your native language to find out what was wrong

ladislav:

PR #89, opened 2013-02-19
help strings of ASSERT and TRANSCODE

ladislav:
bug#1491 related changes, "throws an error" formulation replaced by "causes an error"


BrianHawley:
Nice, it's less confusing and doesn't change the doc string length.

ladislav:

PR #90, opened 2013-02-19
bug#1957 corr, FIRST+ help strings updated

ladislav:
The WORD parameter help string is incompatible with other similar argument help strings (FORSKIP).


ladislav:

PR #91, opened 2013-02-19
bug#1958, BIND and UNBIND help strings corrected

ladislav:
When the BIND (or UNBIND) argument is a block it is not required for it to contain (just) words.


ladislav:

PR #92, opened 2013-02-21
Fix parse thru end

earl:
Commit 44012cc introduced a regression that causes THRU rules to fail when it matches to the end of input. A simple test case illustrating the regression:

```
>> parse "abcd" [thru "d"]
== false  ;; Expected: true
```

The regression was originally reported by Boleslav Brezovsky on AltME.

This fixes [CureCode issue #1959](http://issue.cc/r3/1959).


ladislav:

PR #94, opened 2013-02-22
Fix clipboard device code so it works identically under non-wide-char platforms

ladislav:


ladislav:

PR #98, opened 2013-02-27
DECOMPRESS/GZIP fix

cyphre:
changes details:
-fixed refinement/arg bug in DECOMPRESS
-fixed CRC32 calculation/verification in ZLIB code in DECOMPRESS/GZIP mode

Carl, maybe it would be better to change the /GZIP refinement to /CRC32 ? What do you think?
I think it is better as the feature is not fully compatible with GZIP format but it just uses different checksum type (instead of the default adler-32)
We can always write the higher-level formats (ZIP, GZIP etc.) in REBOL instead of hardcoding into the native code.

### DIFF
--- a/src/core/n-strings.c
+++ b/src/core/n-strings.c
@@ -292,7 +292,7 @@ static struct digest {
 
 	len = Partial1(D_ARG(1), D_ARG(3));
 
-	if (D_REF(4)) limit = Int32s(D_ARG(5), 1);
+	if (D_REF(5)) limit = Int32s(D_ARG(6), 1); // /limit size
 	
 	Set_Binary(D_RET, Decompress(VAL_SERIES(arg), VAL_INDEX(arg), len, limit, D_REF(4))); // /gzip
 
--- a/src/core/s-crc.c
+++ b/src/core/s-crc.c
@@ -374,8 +374,8 @@ static void Make_CRC32_Table(void) {
 	}
 }
 
-static unsigned long Update_CRC32(u32 crc, REBYTE *buf, int len) {
-	u32 c = crc;
+REBCNT Update_CRC32(u32 crc, REBYTE *buf, int len) {
+	u32 c = ~crc;
 	int n;
 
 	if(!crc32_table) Make_CRC32_Table();
@@ -383,7 +383,7 @@ static unsigned long Update_CRC32(u32 crc, REBYTE *buf, int len) {
 	for(n = 0; n < len; n++)
 		c = crc32_table[(c^buf[n])&0xff]^(c>>8);
 
-	return c;
+	return ~c;
 }
 
 /***********************************************************************
@@ -392,7 +392,7 @@ static unsigned long Update_CRC32(u32 crc, REBYTE *buf, int len) {
 /*
 ***********************************************************************/
 {
-	return Update_CRC32(0xffffffffL, buf, len) ^ 0xffffffffL;
+	return Update_CRC32(0x00000000L, buf, len);
 }
 
 
--- a/src/core/u-zlib.c
+++ b/src/core/u-zlib.c
@@ -48,9 +48,9 @@ uLong ZEXPORT adler32(adler, buf, len)
 uLong crc32(uLong num, const Bytef *buf, uInt len)
 {
 #ifndef CRC_DEFINED
-	extern CRC32(unsigned char *, uInt);
+	extern uLong Update_CRC32(uLong, unsigned char *, uInt);
 #endif
-	return num + CRC32((unsigned char *)buf, len);
+	return (len == 0) ? num : Update_CRC32(num, (unsigned char *)buf, len);
 }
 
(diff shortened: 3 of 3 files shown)


BrianHawley:
That refinement change sounds like a good idea to me. Decoding gzip binaries is more involved than just changing the checksum algorithm, but changing the checksum algorithm is important too so the current functionality is still useful, and not just for gzip. A better name will make it less misleading.

cyphre:
Brian, agreed. So this way the COMPRESS/DECOMPRESS format could stay as-is:
[magic-number][compressed-data][checksum][uncopressed-data-size]
where:
magic-number: #{789C}
checksum: adler32 | crc32

I think this format is easy to process at the REBOL level for example in ZIP:// scheme code etc.

PS: maybe we could get rid of the magic-number as well and let the natives add it automatically before passing to ZLIB code. But I don't mind to leave it here for R2 compatibility.

earl:
Changing the refinement effecting this particular behaviour to `/crc32` sounds good to me (or maybe even to `/checksum word`, with `Adler32` and `CRC32` being valid checksum methods).

Besides the obvious `/limit size` handling bug (which I don't think we have a CC report for), this pull request seems to fix the roundtrip issue reported in the comments to [CC#1679](http://issue.cc/r3/1679): with this fix `decompress/gzip compress/gzip "foo"` works as expected.

BrianHawley:
Cyphre, the magic number lets `decompress` detect the compression model without you having to specify it. Most of the time that is how `decompress` is used, without any options specified. Without the magic number, no auto-detection. We might consider using a different magic number for data compressed with CRC32 so R2 doesn't choke on it, though we shouldn't use the zip or gzip numbers except in cases where we use their headers too, and the same goes for other standard compression or archive formats. If we do the gzip or zip support with wrapper code we'll just replace the magic number and headers anyway, so the number only matters to raw `decompress`.

cyphre:
Earl, yes, good points about the refinements.

cyphre:
Brian, that's why I don't care much about the current magic-number. As you said it can be easily removed/replaced in the wrapper code.
The idea to have magic-number per checksum type is good. This way the DECOMPRESS even doesn't have to have the checksum type refinement because by the magic-number the checksum type is auto-detected.

BrianHawley:
Actually, you still want to give `decompress` the option because it would be used by wrappers that use `decompress` but use their own headers and magic numbers. With the option, `decompress` would just do what it's told, regardless of the magic number, and just fail informatively if it was the wrong option. It could even skip past Rebol-specific magic numbers if found (just ours, and we'll have to choose numbers that won't be found at the beginning of deflate-compressed data). That way `decompress` could be used on raw gzip or zip data at the right offset.

Now that I think of it, it might be a good idea for `decompress/crc32` to take a parameter of the CRC32 value that it should check against. Other compression header/footer formats might not put the CRC32 right after the compressed data like `compress` does. Or do they all put it in the same place?

cyphre:
Brian, if I got it right, IMO if we separate the the actual cheksum value into refinement argument then we just move the "problem" at the C code decompressor level (ie the decompressor would need to handle the possible different checksum position).
While if we leave the checksum position fixed as it is now we can always construct this "format" at the REBOL level.(before it is passed to DECOMPRESS)

cyphre:
IIRC the GZIP and ZIP formats have the checksum at the same place. Dunno about others.

BrianHawley:
The decompressor would need to calculate the checksum of the decompressed data while it is decompressing it, but comparing the result to a precalculated checksum to make sure that they're the same is the last step, and not even necessarily a required one if the miniz source and docs are to be believed - miniz also supports raw deflate data with no checksum at all.

The reason that gzip and zip put the checksum at the end is for streaming or incremental block compression. When you don't have the whole dataset until the end of the compression, you don't have the final checksum until then either. Since other deflate-using compression formats are likely to also have that incremental compression constraint, they are also likely to put the checksum in the same place, so there would probably be no problem with `decompress` just assuming that the checksum is at the tail unless it's told otherwise. If PNG puts its checksum somewhere else it wouldn't matter because PNG support is internal, and doesn't call the `decompress` native.

If we want to support having `decompress` not check against a checksum at the tail, we would need a way to say so. That is a good argument for changing the function option style from `decompress/crc32 data` to something like `decompress/check data 'crc32`, because that would let us specify `decompress/check data none` when we want to use a checksum type or position that `decompress` doesn't support yet. The `decompress/check` style is more future-proof too since it lets us add more checksum algorithms later without changing the function options, making `decompress` more `apply` friendly. We'd need to pick a good name for the option though: `/checksum` is precise, but `/check` is more concise and similar to `import/check`.

We can expose some of the lower or higher level capabilities of zlib or miniz as other functions in an internal module (likely named private). That module could be used to implement full zip or gzip support where that's needed.

earl:
Merged to [community](https://github.com/rebolsource/r3).

PR #99, opened 2013-03-06
enhanced EVENT! datatype for event/window and event/gob set-actions

cyphre:
This fix allows to set event/window or event/gob fields. Useful when creating custom GUI events. (make event! [])


ladislav:

PR #102, opened 2013-03-10
BROWSE: Return R_UNSET rather than R_NONE.

0branch:
Following discussion on AltMe, modify `browse` to return `R_UNSET` rather than `R_NONE`.

In addition, handle the `none` case correctly (return early). At present, browsing `none`yields a SEGFAULT; this commit addresses Cure Code ticket #1991 ( http://curecode.org/rebol3/ticket.rsp?id=1991 ).

### DIFF
--- a/src/core/n-io.c
+++ b/src/core/n-io.c
@@ -506,14 +506,16 @@ chk_neg:
 
 	Check_Security(SYM_BROWSE, POL_EXEC, arg);
 
-	if (!IS_NONE(arg))
-		url = Val_Str_To_OS(arg);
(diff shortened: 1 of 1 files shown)


earl:
The unset! change looks good.

NONE handling is slightly contentious: `browse none` currently has a (limited) working functionality on Win32, which would be lost with this change. But if we want to give that functionality up, it'd be better to remove `none!` from BROWSE's argspec completely. If we want that functionality to stay, we'll need a different fix. See the mentioned [ticket #1991](http://issue.cc/r3/1991) for a more detailed discussion.

PR #103, opened 2013-03-11
REWORD API Revamp (See http://issue.cc/r3/1990)

BrianHawley:
Requested changes (by requester):
- Binary support, and working tag support (BrianH)
- `/case` for case-sensitive comparison (BrianH)
- Specifying leading and trailing delimiters with `/escape` (i.e. for HTML/XML entities) (BrianH)
- Being able to specify `none` as an `/escape`, equivalent to `""` now (HostileFork)
- Evaluating value expressions directly in a block spec, rather than requiring the spec to be reduced, and `/only` to turn that off (rgchris, BrianH, prompted by GrahamChiu's RebolBot code)
- Evaluating block values as code like functions, to cut down on function creation overhead, and `/only` to turn that off (rgchris)
- Special-case treatment of set-word keys as if they were regular words, so block and object specs can be compatible (BrianH)
- Special-case treatment of lit-word keys as if they were regular words, to be compatible with fully reduced specs (BrianH)
- Directly using values specs that already match what we need internally, to save overhead (BrianH)
- Better-phrased doc strings that are short enough to not wrap on the Windows console (BrianH)

See http://issue.cc/r3/1990 for the example code.

### DIFF
--- a/src/mezz/mezz-series.r
+++ b/src/mezz/mezz-series.r
@@ -151,50 +151,122 @@ replace: func [
 ; We need to consider adding an /any refinement to use find/any, once that works.
 
 reword: func [
-	"Substitutes values into a template string, returning a new string."
-	source [any-string!] "Template series (or string with escape sequences)"
-	values [map! object! block!] "Pairs of values and replacements (will be called if functions)"
-	/escape "Choose your own escape char (no escape for block templates)"
-	char [char! any-string!] "Use this escape char (default $)"
+	"Make a string or binary based on a template and substitution values."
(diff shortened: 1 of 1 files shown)


BrianHawley:
This was a big group effort, and I made sure every line in those code examples runs perfectly, and that they covered all of the new externally testable features. I also tested the one feature you can't see from the outside, spec reuse.

Once this is accepted I'll update the `reword` article on StackOverflow with the new behavior.

angerangel:
Excellent job Brian!

PR #107, opened 2013-03-21
Add /closure refinement to FUNCT to create closure if needed.

ladislav:
Already used in R3GUI by Cyphre.

### DIFF
--- a/src/mezz/base-funcs.r
+++ b/src/mezz/base-funcs.r
@@ -32,6 +32,7 @@ funct: func [
 	/with {Define or use a persistent object (self)}
 	object [object! block! map!] {The object or spec}
 	/extern words [block!] {These words are not local}
+	/closure {Create a closure}
 ][
 	; Copy the spec and add /local to the end if not found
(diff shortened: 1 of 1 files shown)


BrianHawley:
I recommend that this be rejected and #109 be implemented instead. R3GUI could be easily changed to use `closure` by a simple search and replace, and making things consistent with the `function` renaming of `funct` would have wider community support. Very few people actually use closures yet, and it's even a new thing for R3GUI, and R3 isn't final yet. We would be better off changing things now before neither of these are the case.

No offence is intended, Ladislav, it's a naming thing. API and naming changes like these are the stuff that we need to front-load before too much code is written. Backwards compatibility is only an issue when comparing to existing R2 code at this point - it won't be an issue for R3 code until 3.0 comes out and finalizes the standard.

ladislav:
No problem for me either way, I just need to make sure one of the #109 or this is accepted.

carls:
Closed in favor of PR 109.

PR #109, opened 2013-03-21
FUNCTION and CLOSURE revamp (#1973 and #2002)

BrianHawley:
The function builder name cleanups as requested in http://issue.cc/r3/1973 and http://issue.cc/r3/2002 make the locals-gathering feature of FUNCT more thoroughly integrated into the language, at least as far as its users are concerned. This will give us the advantages of FUNCT without the limits or the readability issue, and make the name FUNCT optional when we start cleaning down a bit.

### DIFF
--- a/src/mezz/base-files.r
+++ b/src/mezz/base-files.r
@@ -157,7 +157,7 @@ split-path: func [
 	reduce [dir pos]
 ]
 
-intern: funct [
+intern: function [
 	"Imports (internalize) words and their values from the lib into the user context."
 	data [block! any-word!] "Word or block of words to be added (deeply)"
 ][
@@ -167,7 +167,7 @@ intern: funct [
 	:data
 ]
 
-load: funct [
+load: function [
 	{Simple load of a file, URL, or string/binary - minimal boot version.}
 	source [file! url! string! binary!]
 	/header  {Includes REBOL header object if present}
--- a/src/mezz/base-funcs.r
+++ b/src/mezz/base-funcs.r
@@ -25,7 +25,7 @@ func: make function! [[
 	make function! reduce [spec body]
 ]]
 
-funct: func [
+function: funct: func [
 	{Defines a function with all set-words as locals.}
 	spec [block!] {Help string (opt) followed by arg words (and opt type and string)}
 	body [block!] {The body block of the function}
--- a/src/mezz/mezz-debug.r
+++ b/src/mezz/mezz-debug.r
@@ -11,7 +11,7 @@ REBOL [
 	}
 ]
 
-dt: delta-time: funct [
+dt: delta-time: function [
 	{Delta-time - returns the time it takes to evaluate the block.}
 	block [block!]
 ][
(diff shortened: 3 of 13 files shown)


BrianHawley:
Had to do this before #107 went through, because `function/closure` just won't fly. The `function` proposal has a lot of community support, and the `closure` change adds the same functionality for closures in a balanced way. This makes locals-gathering seem like the official practice that it has turned out to be in existing R3 code.

The R3GUI project already has a need for locals-gathering closures when making actions. Aside from R3GUI and maybe some Ladislav code, closures aren't used much at all yet. When people see their advantages in R3GUI this will change, and people will start asking why `function` gathers locals and `closure` doesn't.

This change will make R3 code look better, and be easier to debug and maintain too because it is easier, at a glance, to tell the difference between `func` and `function` than it is to distinguish `func` and `funct`. Also, it makes closures seem more on an equal footing with functions, which will come in handy in the future when we have more concurrency.

BrianHawley:
More foolproof, but less efficient in most cases, especially for large function bodies. I'm glad that R3GUI is quietly generating closures for actions, that will likely help, but most of the time people will be better off using functions instead. Closures take all of the function call efficiency gains of R3 and throw them away. They should only be used where it's a good idea to do so. And it is a good idea to do so in some cases, don't doubt that.

BrianHawley:
Ladislav, closures have faster word access, but that is not the efficiency I was talking about. I was talking about memory use. The `bind/copy` of the body block that a closure does every time it is called is definitely bound by the size of the body block and all nested series. Closures trade memory efficiency before the body block is executed for word access speed while the copied body block is being executed. Sometimes this is a good tradeoff, sometimes it isn't.

BrianHawley:
No, it's just not a speed factor. It is definitely an efficiency factor, since memory is another resource that we need to use efficiently. How much memory matters versus speed depends on the circumstances. On desktops, speed matters more.

ladislav:
We may want to reconsider string copying in closure bodies, since it is unnecessary for closures to work.

However, it is true that copying strings in closure bodies makes closures even more user friendly.

In general, the higher friendliness of closures seems to be the least disputed difference between Rebol functions and closures.

ladislav:
Since Carl already dismissed pull request #107 in favour of this one, I am for accepting this pull request ASAP. However, it may make sense to suppress code duplication in some way.

earl:
+1 for merging this pull request soon. For easier integration, I updated (rebased) this pull request to current master and submitted the update as pull #141; this should enable single-click merging.

BrianHawley:
Rebased on current master. #141 can be deleted.

earl:
@BrianHawley Please cherry-pick (or otherwise incorporate) the minor fix from rebolsource/r3@c18c0863a8369a689be7ef077239bd51b5e0245e into this pull request. It's a small change to `make-make.r` so that it continues to work with older R3 builds (that is, builds without the FUNCTION change from this very pull request).

earl:
Pull request #179 combines this pull with the backwards compatibility fix mentioned in my last comment.  Unless this pull request is updated to incorporate this fix as well, please merge #179 instead of this one.

BrianHawley:
Updated with the fix from @earl and rebased. Enjoy!

PR #110, opened 2013-03-25
Collect unused GOBs

ladislav:
Resolving bug#1989.
Define new GOB_MARK and GOB_USED GC flags.
Use the resv field to store GC flags.


ladislav:
This is based on the previous gob/parent handling GC fix.

ladislav:
Tested in Kubuntu Linux with no regressions.
Tested also in Windows, used in a commercial Tree-Mapper application with a good success.

PR #111, opened 2013-03-30
Fix TRIM bug with multiple lines

earl:
This is a slightly reworked version (without the unrelated post-increment to pre-increment changes) of a fix done by @CCX, originally submitted as pull request #76.
## 

Example code exhibiting the bug:

```
>> trim "  x ^/"
== "x x^/"  ;; Expected: "x^/"
```

The fix changes some variable names, to make the algorithm easier to understand. head and tail are now trimmed first, because it is simpler this way.

This fixes [CureCode issue #1948](http://issue.cc/r3/1948).

### DIFF
--- a/src/core/s-trim.c
+++ b/src/core/s-trim.c
@@ -203,8 +203,8 @@ static REBFLG find_in_uni(REBUNI *up, REBINT len, REBUNI c)
 **
 ***********************************************************************/
 {
-	REBCNT start = index;
 	REBCNT out = index;
+	REBOOL append_line_feed = FALSE;
 	REBUNI uc;
 
 	// Skip head lines if required:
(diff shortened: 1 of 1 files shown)


earl:
This supersedes #76.

CCX:
Hum... Its a little chocking (to say the least) seeing a commit signed with my name, specially if I didn't ask to.
Updated my pull request with the suggested —forced— changes. I hope its clean enough to your taste...
TBH, I expected a "Thanks you!" not a "Clean that dirty code!"... specially when it is a fully working fix.
I guess duplicated pull requests are an unnecessary complication. I suggest you delete this duplicate and
keep the original one. (You would want the same treatment for you, wouldn't you?)

earl:
Sorry that we don't yet have written style and contribution guides in place, otherwise this would probably be far more straightforward. I'm of course perfectly fine with keeping the original pull request, as long as we're moving forward there. As for the thanks, please just have a look at the original pull request, which included several thank "Thank you!"s. But thanks again, honestly, for your contribution. Please don't be annoyed by the minor cosmetic refinement requests.

PR #112, opened 2013-04-01
http redirect patch

gchiu:
do-redirect expects an absolute redirect and flags a server error when provided with a relative redirect


gchiu:
Needs to be placed against main

PR #114, opened 2013-04-07
Private module imports take precedence

BrianHawley:
For module Needs imports, give private module exports a
higher priority than runtime library (lib) imports. Doesn't
affect scripts, which use in-order precedence already.

See http://issue.cc/r3/1998 for details.

### DIFF
--- a/src/mezz/sys-base.r
+++ b/src/mezz/sys-base.r
@@ -167,16 +167,16 @@ make-module*: func [
 		; All words of the module body are module variables:
 		bind/new body obj
 		; The module keeps its own variables (not shared with system):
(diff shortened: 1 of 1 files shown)


BrianHawley:
Note that scripts use a different, weirder, precedence model that is due to their shared context being built incrementally. They are in-order precedence for words that were already in the user context before the script is loaded, but were still unset. They are private-module-first for all words exported from the private modules that weren't set in the user context before. Then after all that, any other words that were new in the script and not set by the private module exports ahead of time are added to the user context and initialized from `lib`. Scripts are weird.

PR #135, opened 2013-08-25
Copied DELETE-DIR from Rebol 2 cc#1545

onetom:
Since there was no strong opinions on this
in the http://issue.cc/r3/1545 ticket
I would keep it in Rebol 3 to converge towards
the level of usefulness Rebol 2 provided.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: C
+
+install:
+    - cd make
+    - curl -A '' -O http://www.rebol.com/downloads/v278/rebol-core-278-4-2.tar.gz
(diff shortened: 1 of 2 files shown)


ladislav:
There are three commits unrelated to the DELETE-DIR function in this request, as far as I can see.

onetom:
Well, I did a pull request and only after that I made those 3 more commits, so I don't know why do they appear here... I can just congratulate to Git and Github for their wonderfully obvious to use systems...

Btw, there is a note just above this comment input area saying:
"You can add more commits to this pull request by pushing to the master branch on onetom/rebol"
that's great, but what if i dont want to add more commits to it?
how can i remove commits?

PR #136, opened 2013-08-26
Build Win32 /Core binaries as console mode applications

earl:
A console mode Rebol binary has the following advantages:
- Reuses an existing command prompt
- Works with alternative command prompts (such as Console 2)
- Slightly improves the handling of redirected standard IO

Until we reintegrate /View capabilities and/or a custom graphical (R2-style) console, building console mode binaries is the better default for Windows.


hostilefork:
Big yes vote from me on this change.

ladislav:
I wonder if it isn't better to have a console mode version available in _any_ case, i.e., even when having a custom graphical R2-style console or something else. As I was informed, it might be optimal to have a special "OS_Console" build of the interpreter in addition to View or Core builds.

carls:
It's too bad that Windows still has this duality. I think we'll need both to be supported, as Ladislav noted above.

PR #137, opened 2013-08-26
Copied DELETE-DIR from Rebol 2 (cc#1545)

onetom:
Since there was no strong opinions on this
in the http://issue.cc/r3/1545 ticket
I would keep it in Rebol 3 to converge towards
the level of usefulness Rebol 2 provided.

### DIFF
--- a/src/mezz/base-files.r
+++ b/src/mezz/base-files.r
@@ -124,6 +124,21 @@ make-dir: func [
 	path
 ]
 
+delete-dir: func [
+	{Deletes a directory including all files and subdirectories.}
+	dir [file! url!]
+	/local files
+][
+	if all [
+		dir? dir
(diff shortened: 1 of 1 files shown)


BrianHawley:
This can't be merged as-is. Get rid of the reverse-merge commit.

onetom:
typical git... im not sure how did it happen, that i merge master 2 days later and even push it up.
but i've removed the extra commit, so it should be fine now.

carls:
Originally, I did not bring this forward into R3. I considered that delete of a "file" that's a directory should always be deep, so the native implementation would do that.

There is also the issue that if certain files cannot be deleted, does the function not delete any files?

Anyway, all that said, bringing this forward from Rebol 2 seems fine.

PR #140, opened 2013-08-28
Migrate extensions: *.r => *.reb & *.rx => *.rebx

onetom:
Tested on OS X 10.8.4 like this:
make clean
make prep
make make
make
./r3   # it has booted up successfully

---==| ~/p/rebol-test |==---
$ ~/p/rebol/make/r3 run-recover.r
Testing ...
file: %core-tests.r
recovering at: 4687
Done, see the log file: /Users/onetom/p/rebol-test/r_2_101_0_4_4_C8BB7C_4E5726.log

system/version: 2.101.0.4.4
code-checksum: #{C8BB7C9E4EFD268B5B378F76C67C3DE61969DDD0}
test-checksum: #{4E5726B3CE4645AC5EAB8AF08F677090E0998DD3}
Total: 4696
Succeeded: 4213
Test-failures: 159
Crashes: 16
Dialect-failures: 0
Skipped: 308


earl:
Thanks for working on this, Tamas. Builds successfully on Linux, no test regressions.

I really wonder, though, if that extension change is worth the churn.

onetom:
For the record, the motivation behind this patch is
- differentiate ourselves from the R statistical programming language
- syntax highlighting can simply be based on file extension
- break away from Rebol 2 and don't differentiate Rebol 3, but unite the community over the concept of 1 Rebol
- Carl got convinced about this change: [Let's switch to .reb suffix](http://www.rebol.com/article/0540.html)

anything else I forgot?

ladislav:
I do not know what .rebx is, but maybe it would be preferable to keep the extension at three characters or less?

onetom:
@ladislav It means REBol eXtension. but since it's not used often, I don't think the length matter much. Consistency with the root extension is more important. What's more "we" already have .reds for RED/System; so I'm not bothered at all. the 3 char extension space is very saturated anyway.
Still, I thought about it too for a while of course when i was doing the change, but .rex is an OS/2 script language ext, .rbx is taken by ruby on rails project. Anyway, let's not be so DOS and FAT12 bound!
Also, we can change easily later, since nobody is using it much; unlike .r extension, which we are still changing...

iArnold:
RoR is dead ;)
You cannot be friends with all of the worlds programming history, sometimes you just have to take what belongs to you.

hostilefork:
I can nearly claim success in pushing the logographic hook for Rebol and Red away from believing the most important thing to emphasize is [Rebol starts with R](http://www.rebol.net/w/images/7/72/Rebol-historic-red-program-icon.png) and, [Red starts with R](http://upload.wikimedia.org/wikipedia/commons/4/4b/Logo_R_Redlang.png).

This is a worthwhile change.  Not only is fighting to own the ".r" extension a losing battle in the programming world with all the editors and file extensions, we don't want it.

Even a shortened "rebol" for "rebolution" is a stronger hook than an alphabet block.  R?  R what?  It's the 18th letter of the alphabet, kind of round on top with a leg...too abstract to really have meaning.  Maybe a statistical language is an appropriate fit.  Or maybe for a pirate.  (Arrrrr....)

Either way, let 'em have it.  Of decisions that should be easy, this is the Rebol extension: **.reb**.  If you're from the Rebol 2 world then leave your files as **.r** - don't change 'em.  That can be the distinction.

carls:
Thanks for submitting it. I think this has to happen very soon unless someone shouts "freeze!"

ladislav:
**Clarification request.**

Saphirion currently uses the following "internal extension standard":
- use the **.r** extension for scripts that can be run using both R2 and R3 interpreters
- use **.r3** extension for scripts that can be run using R3
- use **.r2** extension for scripts that can be run using R2

The extensions are visible for public, since Saphirion publishes some of the files (R3-GUI, e.g.)

After accepting this standard, I assume that R3-GUI scripts currently using the **.r3** extension that can run only in R3 should be renamed to use the **.reb** extension.

However, I am unsure how the test-framework files (currently using the **.r** extension and able to run in both R2 and R3) shall be named. A similar question occurs for the **include.r** script, which can also run in both R2 and R3.

onetom:
@ladislav reading your concerns a year later, i still fail to see any practical problem.
The **.reb** extension means "I'm aware of the future and I want to be part of it" and it has nothing to do with the compatibility with the interpreter.
It's a community unification effort which also was supposed to help fading out Rebol 2.

So if I were you, I would rename those **.r** files which run on both Rebol 2 & 3 to **.reb** as well as the **.r3** ones - unless it cause Saphirion some serious problems - and leave the **.r2** ones as is.

But since you technically "shouted Freeze" as @carls put it, I guess I have to update this patch to make it pullable easily.

Do we still want this change?

hostilefork:
@onetom Yes!  Please update the patch so it is easier to pull.

@Ladislav I agree with Tamas that .reb is a fine extension for a Rebol3 file that happens to also be able to run in Rebol2 somehow.  The extension you choose for a file that only runs in Rebol2 can be one of choice.  **.r2** is more informative, but as the world is probably not going to be going around renaming all the old **.r** files to **.r2** in Rebol2 libraries you might also consider putting them in a subdirectory called `/rebol2` and leaving them with the **.r** name.

Though the change has been delayed in the repository itself, the convention has gone forward with the necessary groundwork.  This is the new extension people are using, and **.r3** and **.rebol** are being phased out where they had popped up.  And for those not aware, GitHub now recognizes the .reb extension as Rebol.  _(They've integrated pull requests that improved their syntax highlighting, which is important too.)_

The most important thing is the decision, not the change of some internal files the general public doesn't usually see.  But time to seal the deal!

ladislav:
I surely did not "technically shout freeze". I just asked a question how the extensions are supposed to be used after the change in more complicated situations.. Thanks for your opinion, and to clarify, I do agree that .reb is a fine extension.

onetom:
I'm closing this to clean up my global PR list.

PR #142, opened 2013-08-31
Native 64-bit Linux build

zsx:
This adds native linux 64-bit supports. Tested with github/rebol-test, three progressions and no regression agains master 32-bit found:

``` diff
--- r_2_101_0_4_4_22C633_4E5726.log 2013-08-31 10:01:12.003341307 -0400
+++ r_2_101_0_4_40_183F68_4E5726.log    2013-08-28 09:39:34.563339334 -0400
@@ -4849,13 +4849,13 @@
 [0.1 == round/half-down/to 0.15 0.1] "succeeded"
 [0.2 == round/half-down/to 0.15001 0.1] "succeeded"
 [0.5 == round/half-down/to 0.55 0.1] "succeeded"
-[0.6 == round/half-down/to 0.55001 0.1] "failed"
+[0.6 == round/half-down/to 0.55001 0.1] "succeeded"
 [0.5 == round/half-down/to 0.75 0.5] "succeeded"
 [1.0 == round/half-down/to 0.75001 0.5] "succeeded"
 [-0.1 == round/half-down/to -0.15 0.1] "succeeded"
 [-0.2 == round/half-down/to -0.15001 0.1] "succeeded"
 [-0.5 == round/half-down/to -0.55 0.1] "succeeded"
-[-0.6 == round/half-down/to -0.55001 0.1] "failed"
+[-0.6 == round/half-down/to -0.55001 0.1] "succeeded"
 [-0.5 == round/half-down/to -0.75 0.5] "succeeded"
 [-1.0 == round/half-down/to -0.75001 0.5] "succeeded"
 [0 = sign? 0] "succeeded"
@@ -5825,7 +5825,7 @@
    error? try blk
 ] "succeeded"
 [error? try [apply 'type?/word []]] "succeeded"
-[same? :add attempt [apply does [return/redo :add] []]] "failed"
+[same? :add attempt [apply does [return/redo :add] []]] "succeeded"
 [2 == do does [return apply :do [:add 1 1] 4 4]] "failed"
 [1 == apply :subtract [2 1]] "succeeded"
 [1 == apply :- [2 1]] "succeeded"
@@ -8463,12 +8463,12 @@
 ] "failed, error was caused in the test code"
 [equal? read write clipboard:// c: "test" c] "failed, error was caused in the test code"

-system/version: 2.101.0.4.4
-code-checksum: #{22C633E24BFD4FDBE04A6098E41BD954C3B7D04C}
+system/version: 2.101.0.4.40
+code-checksum: #{183F684918FA91FE87D599821E541D30651FBE9D}
 test-checksum: #{4E5726B3CE4645AC5EAB8AF08F677090E0998DD3}
 Total: 4696
-Succeeded: 4211
-Test-failures: 161
+Succeeded: 4214
+Test-failures: 158
 Crashes: 16
 Dialect-failures: 0
 Skipped: 308
```

### DIFF
--- a/src/core/m-pools.c
+++ b/src/core/m-pools.c
@@ -387,6 +387,7 @@ const REBPOOLSPEC Mem_Pool_Spec[MAX_POOLS] =
 		pool->first = *node;
 		pool->free--;
 		length = pool->wide;
+		memset(node, 0, length);
 	} else {
 		if (powerof2) {
 			// !!! WHO added this and why??? Just use a left shift and mask!
--- a/src/core/m-pools.c
+++ b/src/core/m-pools.c
@@ -639,7 +639,7 @@ clear_header:
 			count++;
 			// The node better belong to one of the pool's segments:
 			for (seg = Mem_Pools[pool_num].segs; seg; seg = seg->next) {
-				if ((int)node > (int)seg && (int)node < (int)seg + (int)seg->size) break;
+				if ((intptr_t)node > (intptr_t)seg && (intptr_t)node < (intptr_t)seg + (intptr_t)seg->size) break;
 			}
 			if (!seg) goto crash;
 			pnode = node; // for debugger
--- a/src/core/m-pools.c
+++ b/src/core/m-pools.c
@@ -420,7 +420,7 @@ const REBPOOLSPEC Mem_Pool_Spec[MAX_POOLS] =
 	memset((REBYTE *)node, 0xff, length);
 #endif
 	series->tail = series->size = 0;
-	SERIES_REST(series) = length / wide;
+	SERIES_REST(series) = length / wide; //FIXME: This is based on the assumption that length is multiple of wide
 	series->data = (REBYTE *)node;
 	series->info = wide; // also clears flags
 	LABEL_SERIES(series, "make");
--- a/src/boot/platforms.r
+++ b/src/boot/platforms.r
@@ -17,7 +17,7 @@ REBOL [
 Amiga		[1 m68k20+ 2 m68k 3 ppc]
 Macintosh	[1 mac-ppc 2 mac-m68k 3 mac-misc 4 osx-ppc 5 osx-x86]
 Windows		[1 win32-x86 2 dec-alpha]
-Linux		[1 libc5-x86 2 libc6-2-3-x86 3 libc6-2-5-x86 4 libc6-2-11-x86 10 libc6-ppc 20 libc6-arm 30 libc6-mips]
+Linux		[1 libc5-x86 2 libc6-2-3-x86 3 libc6-2-5-x86 4 libc6-2-11-x86 10 libc6-ppc 20 libc6-arm 30 libc6-mips 40 libc-x64]
 Haiku		[75 x86-32]
 BSDi		[1 x86]
 FreeBSD		[1 x86 2 elf-x86]
--- a/src/core/a-lib.c
+++ b/src/core/a-lib.c
@@ -708,7 +708,7 @@ RL_API u32 RL_Find_Word(u32 *words, u32 word)
 	return 0;
 }
 
-RL_API int RL_Series(REBSER *series, REBCNT what)
+RL_API uintptr_t RL_Series(REBSER *series, REBCNT what)
 /*
 **	Get series information.
 **
@@ -722,7 +722,7 @@ RL_API int RL_Series(REBSER *series, REBCNT what)
 */
 {
 	switch (what) {
-	case RXI_SER_DATA: return (int)SERIES_DATA(series); // problem for 64 bit !!
+	case RXI_SER_DATA: return (uintptr_t)SERIES_DATA(series);
 	case RXI_SER_TAIL: return SERIES_TAIL(series);
 	case RXI_SER_LEFT: return SERIES_AVAIL(series);
 	case RXI_SER_SIZE: return SERIES_REST(series);
--- a/src/core/a-lib.c
+++ b/src/core/a-lib.c
@@ -104,21 +104,21 @@ extern int Do_Callback(REBSER *obj, u32 name, RXIARG *args, RXIARG *result);
 ***********************************************************************/
 {
 	int marker;
-	REBCNT bounds;
+	uintptr_t bounds;
 
 	Host_Lib = lib;
 
 	if (Host_Lib->size < HOST_LIB_SIZE) return 1;
 	if (((HOST_LIB_VER << 16) + HOST_LIB_SUM) != Host_Lib->ver_sum) return 2;
 
-	bounds = OS_CONFIG(1, 0);
-	if (bounds == 0) bounds = STACK_BOUNDS;
+	bounds = (uintptr_t)OS_CONFIG(1, 0);
+	if (bounds == 0) bounds = (uintptr_t)STACK_BOUNDS;
 
 #ifdef OS_STACK_GROWS_UP
-	Stack_Limit = (REBCNT)(&marker) + bounds;
+	Stack_Limit = (uintptr_t)(&marker) + bounds;
 #else
-	if (bounds > (REBCNT)(&marker)) Stack_Limit = 100;
-	else Stack_Limit = (REBCNT)(&marker) - bounds;
+	if (bounds > (uintptr_t) &marker) Stack_Limit = 100;
+	else Stack_Limit = (uintptr_t)&marker - bounds;
 #endif
 
 	Init_Core(rargs);
--- a/src/core/b-init.c
+++ b/src/core/b-init.c
@@ -96,9 +96,14 @@ extern const REBYTE Str_Banner[];
 #endif
 
 	ASSERT(VAL_TYPE(&val) == 123,  RP_REBVAL_ALIGNMENT);
+#ifdef __LP64__
+	ASSERT(sizeof(REBVAL) == 32,   RP_REBVAL_ALIGNMENT);
+	ASSERT1(sizeof(REBGOB) == 84,  RP_BAD_SIZE);
+#else
 	ASSERT(sizeof(REBVAL) == 16,   RP_REBVAL_ALIGNMENT);
-	ASSERT1(sizeof(REBDAT) == 4,   RP_BAD_SIZE);
 	ASSERT1(sizeof(REBGOB) == 64,  RP_BAD_SIZE);
+#endif
+	ASSERT1(sizeof(REBDAT) == 4,   RP_BAD_SIZE);
 }
 
 
--- a/src/core/b-init.c
+++ b/src/core/b-init.c
@@ -325,7 +325,7 @@ extern const REBYTE Str_Banner[];
 {
 	Action_Count++;
 	if (Action_Count >= A_MAX_ACTION) Crash(RP_ACTION_OVERFLOW);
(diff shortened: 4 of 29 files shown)


ladislav:
Log diff between current master 0.4.4 and this update when the target is 0.4.40 shows one progression here:

```
[10.10.10 == divide 1.1.1 .1] progression, succeeded

Summary:

new-successes: 0
new-failures: 0
new-crashes: 0
progressions: 1
regressions: 0
removed: 0
unchanged: 4695
total: 4696
```

gcc version 4.6.3 (Ubuntu/Linaro 4.6.3-1ubuntu5)

zsx:
I am using gcc (GCC) 4.8.1 20130725 (prerelease) on Archlinux

ladislav:
I used the same compiler for the current master and for the updated code and obtained no difference comparing the current master and this update when the target was 0.4.4 for both.

Thus, the patches are verified to not cause regressions in 0.4.4.

Also verified to not cause regressions in 0.3.1

ladislav:
Also, perhaps interestingly, my http://www.fm.tul.cz/~ladislav/rebol/bench.r runs faster in 0.4.40 than in 0.4.4 although the speed increase is small (+5% to +10%)

earl:
I'm happy to report two more encouraging results: with this series of patches, R3 builds, runs, and tests fine on **Linux on [Itanium (IA64)](https://en.wikipedia.org/wiki/Itanium)** as well as **Linux on [Alpha (AXP)](https://en.wikipedia.org/wiki/DEC_Alpha)**.

onetom:
On a Mac, I got 1 regression:

```
[strict-equal? #[url! ""] to url! ""] crashed
```

```
$ r3 log-diff.r r_2_101_0_4_4_C8BB7C_4E5726.log r_2_101_0_4_40_70D367_4E5726.log
Done.

new-successes: 0
new-failures: 0
new-crashes: 0
progressions: 0
regressions: 1
removed: 0
unchanged: 4695
total: 4696

$ make clean && make make OS_ID=0.4.40 && make prep && time make
...

$ file r3
r3: Mach-O 64-bit executable x86_64
```

onetom:
@ladislav `[strict-equal? #[url! ""] to url! ""] crashed` was the regression (that's what i got when i /usr/bin/diff-ed the filtered test logs, since the log-diff didn't tell it to me as u can see)

not sure what do you mean by "This looks premature", but having this patch would allow us to run rebol on 64bit linux cloud servers without the need of installing the ~100MB ia32-libs dependency.
that alone worth to integrate it, imho, even if there is nothing else 64bit about it...

onetom:
@ladislav okay, it's just bad UX on github; they should emphasize the topic more in the comment header.
i didn't get you were commenting on a specific commit, not on the whole pull request. thanks!

src/core/m-pools.c
```diff
@@ -368,7 +368,11 @@
 
 	CHECK_STACK(&series);
 
+#ifdef __LP64__
+	if (((REBU64)length * wide) > MAX_I64) Trap0(RE_NO_MEMORY);
```
earl: Recovering an [earlier comment of Ladislav](https://github.com/zsx/r3/commit/bebdda231deb8a6b068eb421f33c60912944ee7c#commitcomment-4014963):

"The limit change looks premature, since the implementation of series hasn't changed yet."

zsx: Yes. So you think we should keep it as is until we change the
implementation of series? What I was thinking was that, even with current
series implemention and this patch, you could have a larger series.

Best Regards

Shixin Zeng

ladislav: I think that you obtain a crash trying to allocate a larger series if the limit isn't kept.

zsx: OK, I will remove this commit, even though I still don't quite get where
exactly it could crash due to memory limit.

Best Regards

Shixin Zeng

zsx:
Since I don't have a Mac, I am afraid that I won't be able to fix this.

Best Regards

Shixin Zeng

ladislav:
Zeng, you "lost" the "add a new platform 0.4.40 for linux_x64" commit when updating the pull request.

zsx:
Oops.

Recovered.

Best Regards

Shixin Zeng

src/core/m-pools.c
earl: One example which crashes (crashed) for me:

```
array 134'217'727
```

(With the limit adjustment now removed, this again just causes a "not enough memory" internal error.)

onetom:
@zsx if you give me an ssh public key (practically your github public key), then i could let you in to rebol@mac.rebol.info and you can debug stuff on Mac OS X 10.8.4 Mountain Lion.
u can mail me at onetom at hackerspace sg

zsx:
Thanks @onetom for giving me access to their Mac machine.

Here is what I saw: the code ran fine with r3, but failed in reboltest.

```
mac:rebol-test rebol$ file ../rebol/make/r3
../rebol/make/r3: Mach-O 64-bit executable x86_64
mac:rebol-test rebol$ ../rebol/make/r3 -s
**************************************************************************
**                                                                      **
**  REBOL 3.0 [Alpha Test]                                              **
**                                                                      **
**    Copyright: 2013 REBOL Technologies                                **
**               All rights reserved.                                   **
**    Website:   www.REBOL.com                                          **
**                                                                      **
**    Version:   2.101.0.4.40                                           **
**    Platform:  Linux libc-x64                                         **
**    Build:     11-Sep-2013/15:05:56                                   **
**    Warning:   For testing purposes only. Use at your own risk.       **
**                                                                      **
**    Language:  none                                                   **
**    Locale:    none                                                   **
**    Home:      ../rebol/make/                                         **
**                                                                      **
**************************************************************************

SECURE is disabled

Important notes:

  * Sandbox and security are not available.
  * Direct access to TCP HTTP required (no proxies).
  * Default web browser must be available.

Special functions:

  Chat - open DevBase developer forum/BBS
  Docs - open DocBase document wiki (web)
  Bugs - open CureCode bug database (web)
  Demo - run demo launcher (from rebol.com)
  Help - show built-in help information
  Upgrade - check for newer releases
  Changes - what's new about this version (web)

>> strict-equal? #[url! ""] to url! ""
== true

>>
```

But I do have this line in the log from reboltest:

```
[strict-equal? #[url! ""] to url! ""] "crashed"
```

Am I missing anything?

zsx:
Oops, I closed it by accident. now reopened.

zsx:
Here you go (on the same Mac computer)

```
>> do %catch-any.r
Script: "Catch-any" Version: none Date: none
== make object! [
    do-block: make function! [[
        block [block!]
        exception [word!]
        /local result
    ][
        try [
            catch/quit [
                catch [
                    loop 1 [
                        try [
                            set exception 'return
                            set/any 'result do block
                            set exception none
                            return :result
                        ]
                        set except...

>> catch-any [strict-equal? #[url! ""] to url! ""] 'exception
== true

>> exception
== none
```

zsx:
Then something broken in reboltest?

zsx:
Here's the result:

```
mac:rebol-test rebol$ ../rebol/make/r3 -s run-recover.r 
Testing ...
file: %core-tests.r
new log
Done, see the log file: /Users/rebol/rebol-test/r_2_101_0_4_40_B4877A_827016.log

system/version: 2.101.0.4.40
code-checksum: #{B4877AF33A0B0CDE9E1879C5A7191CA11BB674F0}
test-checksum: #{827016D57EF0B0055F786EF539D658CD340CA054}
Total: 1
Succeeded: 1
Test-failures: 0
Crashes: 0
Dialect-failures: 0
Skipped: 0

mac:rebol-test rebol$ cat core-tests.r 
; datatypes/action.r
[strict-equal? #[url! ""] to url! ""]
```

earl:
When I comment out the `[strict-equal? #[url! ""] to url! ""]` test, the test suite fails with a "log file parsing problem" error. Here's the tail of the logfile:

```
[unset? to unset! none] "succeeded"
[unset? to unset! 1] "succeeded"
[error? try [a: ()]] "succeeded"
[not error? try [set/any 'a ()]] "succeeded"
[url? http://www.fm.tul.cz/~ladislav/rebol] "succeeded"
[not url? 1] "succeeded"
[url! = type? http://www.fm.tul.cz/~ladislav/rebol] "succeeded"
[url? #[url! ""]] "succeeded"
[strict-equal? #[url! ""] make url! 0] "succeeded"
["http://" = mold http://] "succeeded"
```

However, a plain 32-bit build parses and recovers from this logfile just fine. (Also, a 32-bit build runs the test suite with the commented-out test without problems.)

To me, this, symptomatically, looks like a memory corruption issue lingering somewhere.

earl:
After some debugging together with Ladislav, it seems we found the cause. The following crashes 64-bit OSX builds (during GC):

```
foo: make module! [[] []]
recycle
```

(The full test in the test framework that causes this is in [lines 1824-1834](https://github.com/rebolsource/rebol-test/blob/c7df0ba63a8e556a0c6b4d8ed00495c7386d49de/core-tests.r#L1824-L1834).)

Here's the full diff of regressions after the latest test framework adaptations (which reveal GC bugs such as this more precisely): https://gist.github.com/earl/1f22142b5e8d9b521376

zsx:
``` ​
Also, LP64 seems to make the REBALL type definition incorrect.
```

Could you elaborate on this?

ladislav:
In the sys-value.h header file, REBALL is described as "All bits of value fields:". The REBALL type definition is not compatible with that when **LP64** is defined. However, as I checked, the discrepancy does not cause any harm except for making the description invalid.

zsx:
REBALL is easy to be fixed. We just need to change it to uintptr_t bits[3].

ladislav:
Well, it certainly isn't (too) hard to fix, but this fix might break s-crc.c, I am afraid.

ladislav:
Also, #147 is worth consulting, since it both suppresses compiler warnings for 32-bit cases and can do the same service for 64-bit case where applicable.

ladislav:
This looks unbelievable, but 32-bit R3 actually became slower (less than 5%) after your last batch of changes.

While speed of 64-bit R3 did not change noticeably.

zsx:
Compared with the https://github.com/rebol/rebol master?

Best Regards

Shixin Zeng

ladislav:
No, I just had your 64-master state recorded before you made the last batch of changes including struct rearrangement and I compared speed of the older 64-master with the speed of the new 64-master containing your last batch of changes including struct rearrangement.

Best Regards, Ladislav

ladislav:
I guess I know why the struct rearrangement does not have positive effect: there are the #pragma pack(4) directives preceding many struct declarations and they were most probably meant for 32-bit programs. Curiously enough, I guess that they don't have any effect on 32-bit programs.

zsx:
I think that should actually be the reason it should improve the
performance. When it's packed at 4, 64-bit types might not aligned with
their natural boundaries. With the struct rearrangement, these types should
be aligned to their natural boundaries and still packed at 4.

Best Regards

Shixin Zeng

ladislav:
Aha, well, but there was no performance improvement as I noted above, and a slow down in 32-bit version, so this is not what you expected.

Best Regards

Ladislav

zsx:
Yep, that's kind of surprising. I'll take a look when I have time.

Best Regards

Shixin Zeng

carls:
See newer #178 pull for latest 64 bit support. Closing this pull.

PR #128, opened 2013-08-15
Use re-binding for closure bodies (cc#2048)

ladislav:
Based on pull #123. Fixes [CC#2048](http://issue.cc/r3/2048). Tested in Linux (0.4.4).

PR #123, opened 2013-08-05
Let BIND bind out-of scope function words (cc#1893)

ladislav:
Based on pull #125. Fixes [CureCode issue #1893](http://issue.cc/r3/1893). Tested in Linux (0.4.4).

PR #122, opened 2013-08-04
Fix function frame identification (cc#2025)

ladislav:
This fixes [CureCode issue #2025](http://issue.cc/r3/2025) and [CureCode issue #2044](http://issue.cc/r3/2044).Tested in Linux (0.4.4).

PR #143, opened 2013-08-31
Remove SAVE_SERIES/UNSAVE_SERIES around code not calling Recycle.

ladislav:
Based on #139 (because #139 includes commit 7f691b06 which removes two other unneeded SAVE_SERIES/UNSAVE_SERIES pairs) . Tested in Linux (0.4.4).


carls:
Not sure how these get back referenced by pull number, so adding this comment.
More details found in these closed pulls: #122, #123, #125, #128, #132, #133, #134, #139

earl:
A link to other pulls is automatically created if you use a `#` prefix (so `#122` renders as #122).

carls:
Lots of bind changes here, but 1) they're coming from Ladislav (worldwide bindology expert), and 2) tested by community build for many months.

Also note that the title of this pull doesn't indicate it's an aggregate that includes all of the other related pulls. See list above.

PR #145, opened 2013-09-08
Modify INTEGER! multiplication to not use division for overflow detection.

ladislav:
Also, the modified multiplication does not use _undefined arithmetic_.
Modify INTEGER! addition and subtraction to not use _undefined arithmetic_ and optimize.
Modify INTEGER! comparison to not use _undefined arithmetic_ - fix [bug#2054](http://issue.cc/r3/2054)
Do not use MAX_NEG_INT (problematic name).
Amend a comment referring to reb-config.h
Tests in Linux (0.4.4) reveal 13 progressions and no regression
Multiplication speed tests reveal improvement

### DIFF
--- a/src/core/t-integer.c
+++ b/src/core/t-integer.c
@@ -37,10 +37,9 @@
 /*
 ***********************************************************************/
 {
-	REBI64 num = VAL_INT64(a) - VAL_INT64(b);
-	if (mode >= 0)  return (num == 0);
-	if (mode == -1) return (num >= 0);
-	return (num > 0);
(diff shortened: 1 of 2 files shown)


earl:
Tested successfully on Linux-x86: no regressions, 13 progressions.

ladislav:
This commit is also meant to replace the [64-master crashing fix](https://github.com/zsx/r3/commit/df4533621d0c6180c0a123a14b07ec6a21eeedcc) commit

PR #147, opened 2013-09-19
Suppress compiler warnings.

ladislav:
Delete REBDCL type definition (no longer used in code).
Tested in Linux (0.4.4), no regressions
Based on #145 since it updates reb-c.h

### DIFF
--- a/src/core/f-deci.c
+++ b/src/core/f-deci.c
@@ -536,8 +536,8 @@ REBDEC deci_to_decimal (const deci a) {
 deci decimal_to_deci (REBDEC a) {
 	deci result;
 	REBI64 d; /* decimal significand */
-	REBINT e; /* decimal exponent */
-	REBINT s; /* sign */
+	int e; /* decimal exponent */
+	int s; /* sign */
 	REBYTE *c;
 	REBYTE *rve;
 
--- a/src/core/f-math.c
+++ b/src/core/f-math.c
@@ -392,7 +392,8 @@ static int Convert_Decimal(REBDEC d, REBI64 *sig, REBINT *point)
 
 REBINT Emit_Decimal(REBYTE *cp, REBDEC d, REBFLG trim, REBYTE point, REBINT decimal_digits) {
 	REBYTE *start = cp, *sig, *rve;
-	REBINT e, sgn, digits_obtained;
+	int e, sgn;
+	REBINT digits_obtained;
 
 	/* sanity checks */
 	if (decimal_digits < MIN_DIGITS) decimal_digits = MIN_DIGITS;
--- a/src/core/f-round.c
+++ b/src/core/f-round.c
@@ -86,7 +86,7 @@ enum {
 ***********************************************************************/
 {
 	REBDEC r;
-	REBINT e;
+	int e;
 	REBFLG v;
 	union {REBDEC d; REBI64 i;} m;
 	REBI64 j;
--- a/src/core/u-gif.c
+++ b/src/core/u-gif.c
@@ -294,7 +294,7 @@ void Chrom_Key_Alpha(REBVAL *v,REBCNT col,REBINT blitmode) {
 			Append_Series(VAL_SERIES(Temp2_Value), (REBMEM *)Temp_Value, 1);
 		}
 */
-		dp = codi->bits = Make_Mem(w * h * 4);
+		dp = codi->bits = (u32 *)Make_Mem(w * h * 4);
 		codi->w = w;
 		codi->h = h;
 
--- a/src/core/u-jpg.c
+++ b/src/core/u-jpg.c
@@ -1,6 +1,7 @@
 #define JPEG_INTERNALS
 #define NO_GETENV
-
+#include "reb-config.h"
+#include "reb-c.h"
 #include <setjmp.h>
 #include "sys-jpg.h"
 
@@ -10788,7 +10789,7 @@ extern void Register_Codec(char *name, codo dispatcher);
 
 /***********************************************************************
 **
-*/	int Codec_JPEG_Image(REBCDI *codi)
+*/	REBINT Codec_JPEG_Image(REBCDI *codi)
 /*
 ***********************************************************************/
 {
@@ -10810,7 +10811,7 @@ extern void Register_Codec(char *name, codo dispatcher);
 	if (codi->action == CODI_DECODE) {
 		int w, h;
 		jpeg_info(codi->data, codi->len, &w, &h);
-		codi->bits = (unsigned int *)Make_Mem(w * h * 4);
+		codi->bits = (u32 *)Make_Mem(w * h * 4);
 		jpeg_load(codi->data, codi->len, (char *)codi->bits);
 		codi->w = w;
 		codi->h = h;
--- a/src/include/reb-c.h
+++ b/src/include/reb-c.h
@@ -21,12 +21,11 @@
 **
 **  Summary: General C definitions and constants
 **  Module:  reb-c.h
-**  Author:  Carl Sassenrath
+**  Author:  Carl Sassenrath, Ladislav Mecir
 **  Notes:
 **      Various configuration defines (from reb-config.h):
 **
 **      HAS_LL_CONSTS - compiler allows 1234LL constants
-**      INT_64_MODE - int size is 64 bits
 **      ODD_INT_64 - old MSVC typedef for 64 bit int
 **      OS_WIDE_CHAR - the OS uses wide chars (not UTF-8)
 **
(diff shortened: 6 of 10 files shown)


ladislav:
Merged to community

earl:
As a note to anyone else reviewing this, there's one non-obvious change as part of introducing the (optional) usage of C99 integer types.

[Previously](https://github.com/rebol/rebol/pull/147/files#diff-6ef5cfafbcfde94617f41a68573fbe68L57), `i32` and `u32` where defined via `long`:

```
typedef long            i32;
typedef unsigned long   u32;
```

[Now](https://github.com/rebol/rebol/pull/147/files#diff-6ef5cfafbcfde94617f41a68573fbe68R76), these types are either defined via C99's `[u]int32_t` or they are defined via `int` (when C99 isn't available):

```
typedef int            i32;
typedef unsigned int   u32;
```

This is a desired change.

On 32-bit systems with an ILP32 data model (which includes all 32-bit targets we currently support), these definitions are equivalent (as far as storage size is concerned): both `long` and `int` are 32 bit wide with ILP32.

On 64-bit systems with a LP64 (I32LP64) data model (such as 64-bit Linux), the old typedef was actually incorrect. `long` is 64 bits wide with  LP64; the new typedef is correct for LP64.

On 64-bit systems with a LLP64 (IL32P64) data model (such as 64-bit Windows), the old and the new typedef are again equivalent regarding storage size: `int` and `long` are 32 bits wide with LLP64.

We currently don't (and probably won't for the foreseeable future) support any 16-bit, 32-bit non-ILP32 or 64-bit [S]ILP64 systems. (And in any case, neither the old nor the new definitions will have the desired effect for those systems.)

hostilefork:
@carls asked for [feedback on this in chat](http://chat.stackoverflow.com/transcript/message/14744314#14744314):

> Request your feedback and consideration of [this PR]
> My primary concern is what might happen on systems with older compilers.

**Short Feedback**

Rather than making decisions based on hypothetical older compilers, we should go ahead and do what it takes to get emulators (or someone with a physical machine)... with actual copies of the compilers we may be concerned about.  See what it takes to make things work... decide if it's worth keeping it working... and add to the "wall of binaries" (and possibly automated builds).

I think Amiga is a good example, for the historical/museum aspects of it.  Consider that some Googler thinks it's worth it to emulate an Amiga 500 as a demo of the Native Client, and it gets written up in Ars Technica:

http://arstechnica.com/information-technology/2013/12/run-the-amiga-500-in-your-browser-with-portable-native-client/

How about loading that up and running Rebol 3 inside the NaCl emulator in the browser, with modified semantics for FUNCTION and all?  Wouldn't that be something?

I know time is limited, but as @DocKimbel says: "programming is supposed to be fun".  Wouldn't that be an intersection between fun and this kind of question?  Shouldn't we be deciding what changes we accept based on a cool demo baseline worth preserving (for the art of it, if nothing else) instead of hunches?

So can we define a couple of the oldest compilers to accept as standard--and get working copies of those compilers?  Then we can assume if master branch can be kept working on those more or less, that we're not drifting too far from a stable baseline.  I'll suggest the Amiga as a stake in the ground, of personal relevance to @carls and others, and a good benchmark of _"well it's still working there, so the change must not be too disrupitive"_

**Much Longer Feedback**

@earl noted in the previous message: _"We currently don't (and probably won't for the foreseeable future) support any 16-bit, 32-bit non-ILP32 or 64-bit [S]ILP64 systems. (And in any case, neither the old nor the new definitions will have the desired effect for those systems.)"_

[@iArnold responded by saying](http://chat.stackoverflow.com/transcript/message/14744482#14744482):

> So follow the direction of time, forward.

To the extent I understand the arrow of time, physics doesn't seem to think it very foundational to its properties.  All except for the "direction of time" being defined by increasing entropy, that is.  So maybe following time unconditionally is a bad idea.  :-)

Part of the philosophy of Rebol is to show an alternative to the grisly progress of unchecked evolution.  That's not to say that the endless oceans of evolution should be ignored--you can catch interesting new  fish there.  Yet when you're out exploring the sea, you recognize the value of having an anchor on the boat.

As a meditative point from [my homeboy Don Knuth](http://laughingsquid.com/jacob-appelbaum-donald-knuth-demonstrate-the-recursive-homeboys-principle/), this is what Wikipedia says about the versioning system for TeX and Metafont:

> "Since version 3, TeX has used an idiosyncratic version numbering system, where updates have been indicated by adding an extra digit at the end of the decimal, so that the version number asymptotically approaches π. This is a reflection of the fact that TeX is now very stable, and only minor updates are anticipated. The current version of TeX is 3.14159265; it was last updated 2014-01-12.[10] The design was frozen after version 3.0, and no new feature or fundamental change will be added, so all newer versions will contain only bug fixes. Even though Donald Knuth himself has suggested a few areas in which TeX could have been improved, he indicated that he firmly believes that having an unchanged system that will produce the same output now and in the future is more important than introducing new features. For this reason, he has stated that the "absolutely final change (to be made after my death)" will be to change the version number to π, at which point all remaining bugs will become features.[11] Likewise, versions of METAFONT after 2.0 asymptotically approach e, and a similar change will be applied after Knuth's death."

I see Rebol as best targeting that desire for a hardened, thought-out design...like TeX.  It's at an earlier phase of evolution; only open-source for a year, so the peer review hasn't been strong as it needs to be.  Also, there's a statistically insignificant number of Rebol 3 codebases, which shouldn't hold back a good change for the sake of legacy.  There are really still a lot of foundational questions, like:
- Should `if <foo> = "FoO" [print "="]` result in printing "="?
- Is it a good idea to support `if/else 1 < 2 [print "<"] [print ">="]`
- Should you have to write `rejoin ["foo" either has-bar? [{bar}] [{}] "baz"]` or can you just write `rejoin ["foo" (if has-bar? {bar}) "baz"]`

Such "tiny" questions only scratch the surface of how much guru meditation there is to do.  In CureCode and R3 DevBase there are many unresolved points swimming in the ["Deep Lake"](http://www.rebol.com/article/0103.html).  If Rebol/Red matter to the future at all, then it matters to become efficient about using modern workflows to identify and triage the collective knowledge that has already been gathered.

Red likes to market its lack of dependence on a standard toolchain.  Which is great, but calls out the contrasting marketing point for Rebol to an audience with different requirements: _it uses only a standard ANSI-C toolchain_^^.  I think the best way to chart forward is to say that it always will do so...and always be written in ANSI C.  It means the tool can be rapidly ported to nearly any platform.

^^_(Although I do like the idea of being --able-- to build it with a C++ compiler.  So hopefully we can progress and make that happen; I had it working a year ago.  Once it does, that opens building with some C++-only compilers... some optimizations... and the ability to use C++ features and type system in testing things.  To pick a random example: strongly typed enums.)_

Bringing it back to the question: instead of asking _"why do you care if you can build Rebol on an obscure system?"_ I think we should contextualize this as _"why SHOULDN'T you be able to build Rebol on an obscure system?"_  If Rebol can still be built on Amigas and Haiku then why not?  What about a VAX?  There are obvious applications to digital archiving operations... and people are passionate enough about it to make emulators:

http://en.wikipedia.org/wiki/List_of_computer_system_emulators#32-bit_systems

Plus having a "wall of binaries" to pull from is an opportunity to show one of the great Rebol strengths.  It would be nice to have someone go ahead and put together an Amiga build, and make a YouTube video of it running.  Keeping that "it just works" is--if it does nothing else--reminding people that they _could_ be working that way.

More importantly: my overarching belief is that Red should hold back on bootstrap as long as possible, and even after bootstrapping be able to re-bootstrap using Rebol 3.  Which means any platform Rebol runs on with a standard toolchain breaks ground to let Red run on that platform to cross-compile for any platform it supports (even if it doesn't yet support the platform its running on).

_(Note: I actually got pretty far in a [Red Rebol 3 port](https://github.com/red/red/pull/421/files) to build a binary-equivalent `hello.red` on Linux.  But the questions it raised about language design decisions and Rebol2/Rebol3 changes really drove back to why we need to focus on getting consensus on fundamentals...which is where I want the whole Rebol effort to turn its laser sights on.)_

Let Red chase the evolutionary ocean, and Rebol be the anchor.  And let's anchor Rebol to some _specific_ old platforms...and get those builds working and downloadable onto the machine in your closet or the emulator of choice.

And though I want to focus Rebol's development track onto a full triage of foundational language issues (instead of multicore parallelism, OpenCV integration, or using native Android widgets)... 64-bit builds are a very good target.  Precisely because it is so relevant to the basic issues of TO BINARY! conversions of integer and other things that we might futureproof as part of the specification, such as this which I dismissed in a bout of irritation... but still believe is the kind of thinking we need to be having:

http://curecode.org/rebol3/ticket.rsp?id=2064&cursor=647

ladislav:
I updated this pull request keeping the 

 typedef long            i32;
 typedef unsigned long   u32;

definitions for non-C99 and non-LP64 compilers as before.

carls:
Having used dozens of compilers that are problematic in these definitions, I tend to get concerned... mainly because I'm the one that ends up fixing them. However, let's go ahead with this change, and when someone complains that they cannot compile the source, let's not say to them "sorry, you must update everything," but rather, let's fix it -- probably a minor change here.

PR #148, opened 2013-09-23
Remove int32 from Reb_Val_Data (not used)

ladislav:
Tested in Linux (0.4.4) with no regressions.


ladislav:
Merged to community.

PR #149, opened 2013-09-24
Fix SHIFT CC#2067

ladislav:
Replace _undefined arithmetic_
Detect overflow of arithmetic shift
Fix [CC#2067](http://issue.cc/r3/2067)
Tested in Linux (0.4.4) with 48 progressions and no regression.


ladislav:
Merged to community.

PR #150, opened 2013-09-25
Fix circular block compare crash, CC#1049

ladislav:
Fixes [CC#1049](http://issue.cc/r3/1049)
Tested in Linux (0.4.4) with 3 progressions (successes instead of crashes) and no regression


ladislav:
Merged to community.

PR #151, opened 2013-09-26
Fix EVEN? and ODD? to not have undefined behaviour

ladislav:
Make EVEN? and ODD? result tolerant to differences from integral values smaller than 0.5
(the greatest tolerance possible in this case)
Tested in Linux (0.4.4) with no regressions.
Tests in Adroid reveal progressions and no regression.

The original version was using an expression which could overflow and thus have _undefined behaviour_. Also, the original version did not tolerate even the smallest differences from integral values, see [CC#1622](http://issue.cc/r3/1622).


ladislav:
Merged to community.

PR #152, opened 2013-09-26
Use Bentley & McIlroy's qsort for sorting strings and blocks

ladislav:
Currently, R3 uses platform-specific code for sorting.
This may increase the effort necessary to port the interpreter to new platforms.

This implementation brings the following advantages:
- The sorting algorithm is stable fixing [CC#1152](http://issue.cc/r3/1152)
- Sort results will be consistent across wide range of platforms
- Makes porting to new platforms easier
- Battle-tested solution, as this is the way how the problem was solved in e.g. OpenBSD, FreeBSD, Apple OS, Android

The commit makes the following changes:
- Bentley & McIlroy's qsort is added as src/core/f-qsort.c.
- The source is unmodified except for a reference to where the source was downloaded from and two parts marked "commented-out"
- An attribution notice is added to NOTICE.
- The additional source is listed in src/tools/file-base.r
- Linux (0.4.4) makefile was regenerated to build the new f-qsort.r

Tested in Windows with no regression and 5 progressions.
Tested in Linux (0.4.4) with no regression.


earl:
Please also re-generate the default bundled (Linux/x86, 0.4.4) makefile, to have it build the new `f-qsort.c`.

ladislav:
Done.

ladislav:
Merged to community.

PR #153, opened 2013-09-27
Prevent word duplication when appending to an object

ladislav:
- Fix [CC#1979](http://issue.cc/r3/1979)
- Amend 'self handling
- Fix [CC#2076](http://issue.cc/r3/2076)
- Optimize the code to not use search

Based on #171 to not cause crashes during the startup of 64-bit R3.

Tested in Linux (0.4.4), no regressions, 4 progressions.

### DIFF
--- a/src/core/c-frame.c
+++ b/src/core/c-frame.c
@@ -265,11 +265,13 @@
 **
 ***********************************************************************/
 {
-	REBVAL *words;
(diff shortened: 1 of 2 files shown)


ladislav:
This fix replaces https://github.com/rebol/rebol/pull/105

ladislav:
Merged into community.

PR #154, opened 2013-09-30
Fixes warnings when building & terminal support under Haiku

Tectorum:
Haiku's terminal uses ESC-O instead of ESC-[ as described [here](http://www.freelists.org/post/haiku/Nonstandard-CSI-code,5) in part to strictly comply with xterm. In addition, it defines unsigned ints but doesn't define DEF_UINT in its headers, which caused warning when building. These changes enable additional console functionality and remove the warnings when building.

### DIFF
--- a/src/include/reb-config.h
+++ b/src/include/reb-config.h
@@ -140,6 +140,7 @@ These are now obsolete (as of A107) and should be removed:
 #define API_EXPORT __attribute__((visibility("default")))
 #else
 #define API_EXPORT
+#define DEF_UINT
 #endif 
 
 #define API_IMPORT
--- a/src/os/posix/host-readline.c
+++ b/src/os/posix/host-readline.c
@@ -429,7 +429,7 @@ static struct termios Term_Attrs;	// Initial settings, restored on exit
 	if (*cp == ESC) {
 		// Escape sequence:
 		cp++;
-		if (*cp == '[') {
+		if (*cp == '[' || *cp == 'O') {
 
 			// Special key:
 			switch (*++cp) {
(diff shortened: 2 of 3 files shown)


hostilefork:
Hello Tectorum!  Great to see someone from Haiku taking a look!  _(I did the minor modifications needed for the first build, and made [a video about that](http://www.youtube.com/watch?v=jImeByEkV8Y), if you didn't see it)_

The console support in Rebol3 is pretty bad right now on all platforms, and definitely something we want to see improved.  We're still setting up some of the organization regarding how pull requests are migrated into "stable" vs "development".  So it might be a little bit before this gets pulled in, but when I get a second to make another Haiku VM I'll chime in with a "yup, works" :-)

In the meantime, please feel free to come talk with us in the [Rebol and Red StackOverflow chat](http://rebolsource.net/go/chat-faq).  I've been wanting to figure out what we'd have to do to get Rebol properly boxed up for Haiku's package management, and know pretty much nothing about that... or also, if there are any pieces of the Haiku infrastructure that Rebol might be able to be used in.

pierrechtux:
Hello!

I have always been bothered by the ESC-[ in Rebol: it caused me confusion in square brackets matching, and also when I define foldmarker=[] in vim
So, if a different convention could be adopted, for all platforms, I'd be quite happy.

Tectorum:
Hi Hostilefork, and thanks! I've just started working with Haiku and Rebol, and your video was very helpful in getting started. I'll check out the StackOverflow chat & Haiku's package management later in the week.

Hi Pierrechtux! You may be able to change termcap to use ESC-O as a temporary solution, but I think ESC-[ is the [modern standard](http://www.ecma-international.org/publications/standards/Ecma-048.htm).

ladislav:
Tested in 0.4.4, no regressions.

ladislav:
Can somebody run the tests in Haiku, please?

earl:
Tested on Haiku-x86 (0.5.75), no regressions.

PR #155, opened 2013-09-30
Add ZLIB subsetting script and update snapshots to latest (1.2.8)

hostilefork:
This creates a new make target:

```
make zlib
```

When you run that, it will execute the script **%src/tools/make-zlib.r**.  This will process the [Zlib repository](https://github.com/madler/zlib) to produce two snapshot files that Rebol requires: **%src/include/sys-zlib.h** and **%src/core/u-zlib.c**.

_(Note: Until HTTPS support is in mainline, you will have to download or clone the Zlib sources locally and edit the path in the script to point to wherever you put it.)_

This snapshotting process would be manually re-run once by someone after any new release of Zlib.  The subsetting script would need to be updated if there were any significant changes to the library's structure.

This appears to resolve [CC #2040](http://curecode.org/rebol3/ticket.rsp?id=2040), as with the change I can now successfully save PNG files on Linux.

### DIFF
--- a/src/core/u-compress.c
+++ b/src/core/u-compress.c
@@ -71,6 +71,12 @@
 **
 ***********************************************************************/
 {
+	// NOTE: The use_crc flag is not present in Zlib 1.2.8
+	// Instead, compress's fifth paramter is the compression level
+	// It can be a value from 1 to 9, or Z_DEFAULT_COMPRESSION if you
+	// want it to pick what the library author considers the "worth it"
+	// tradeoff of time to generally suggest.
+
 	REBCNT size;
 	REBSER *output;
 	REBINT err;
@@ -82,7 +88,7 @@
 
 	//DISABLE_GC;	// !!! why??
 	// dest, dest-len, src, src-len, level
-	err = Z_compress2(BIN_HEAD(output), (uLongf*)&size, BIN_HEAD(input) + index, len, use_crc);
+	err = z_compress2(BIN_HEAD(output), (uLongf*)&size, BIN_HEAD(input) + index, len, Z_DEFAULT_COMPRESSION);
 	if (err) {
 		if (err == Z_MEM_ERROR) Trap0(RE_NO_MEMORY);
 		SET_INTEGER(DS_RETURN, err);
@@ -108,6 +114,9 @@
 **
 ***********************************************************************/
 {
+	// NOTE: The use_crc flag is not present in Zlib 1.2.8
+	// There is no fifth parameter to uncompress matching the fifth to compress
+
 	REBCNT size;
 	REBSER *output;
 	REBINT err;
@@ -123,7 +132,7 @@
 	output = Make_Binary(size + 20); // (Why 20 extra? -CS)
 
 	//DISABLE_GC;
-	err = Z_uncompress(BIN_HEAD(output), (uLongf*)&size, BIN_HEAD(input) + index, len, use_crc);
+	err = z_uncompress(BIN_HEAD(output), (uLongf*)&size, BIN_HEAD(input) + index, len);
 	if (err) {
 		if (PG_Boot_Phase < 2) return 0;
 		if (err == Z_MEM_ERROR) Trap0(RE_NO_MEMORY);
--- a/src/core/u-zlib.c
+++ b/src/core/u-zlib.c
@@ -1,62 +1,1097 @@
-#include "sys-zlib.h"
-#include <stdlib.h>
+/***********************************************************************
+**
+**  Extraction of ZLIB compression and decompression routines
+**  for REBOL [R3] Language Interpreter and Run-time Environment
+**  This is a code-generated file.
+**
+**  ZLIB Copyright notice:
+**
+**    (C) 1995-2013 Jean-loup Gailly and Mark Adler
+**
+**    This software is provided 'as-is', without any express or implied
+**    warranty.  In no event will the authors be held liable for any damages
+**    arising from the use of this software.
+**
+**    Permission is granted to anyone to use this software for any purpose,
+**    including commercial applications, and to alter it and redistribute it
+**    freely, subject to the following restrictions:
+**
+**    1. The origin of this software must not be misrepresented; you must not
+**       claim that you wrote the original software. If you use this software
+**       in a product, an acknowledgment in the product documentation would be
+**       appreciated but is not required.
+**    2. Altered source versions must be plainly marked as such, and must not be
+**       misrepresented as being the original software.
+**    3. This notice may not be removed or altered from any source distribution.
+**
+**        Jean-loup Gailly        Mark Adler
+**        jloup@gzip.org          madler@alumni.caltech.edu
+**
+**  REBOL is a trademark of REBOL Technologies
+**  Licensed under the Apache License, Version 2.0
+**
+************************************************************************
+**
+**  Title: ZLIB aggregated source file
+**  Build: A0
+**  Date:  29-Sep-2013
+**  File:  u-zlib.c
+**
+**  AUTO-GENERATED FILE - Do not modify. (From: make-zlib.r)
+**
+***********************************************************************/
 
-#define BASE 65521L /* largest prime smaller than 65536 */
+#include "sys-zlib.h" // added by make-zlib.r
+
+/* crc32.c -- compute the CRC-32 of a data stream
+ * Copyright (C) 1995-2006, 2010, 2011, 2012 Mark Adler
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ *
+ * Thanks to Rodney Brown <rbrown64@csc.com.au> for his contribution of faster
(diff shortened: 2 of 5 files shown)


earl:
Tested on Linux/x86: no regressions, 1 progression (the test for cc#2040).

carls:
Generally, I like finding a good lib release and sticking with it. By good lib, I mean one that does 99% of what you need and is small and well tuned. The original zlib was that way. Served well and quite stable over a long period.

It seems that all good designs peak and then head downhill after that, both in terms of reliability, security, efficiency and size. I don't subscribe to the notion that newer is better. Often it's not. (E.g. when I find a good distro of Linux, I stick with it for many years and don't upgrade just for upgrade's sake.)

That said, the current zlib is quite old. It's done a good job over 15 years, but yes, it's probably time to replace it. The problem is, I don't know about this newer version. Is it faster, smaller, better? Does it include a lot of features we'll never use? Anyone know?

hostilefork:
The size delta question of the update will of course vary from compiler to compiler.  But for the ballpark:  a release core build on Win32 before the update came out as 438,784.  After the update, without doing any configuring or investigation, it's 457,728.  Net delta of 18,944 bytes, so 18.5K.

The copyright on the previous code is to 1998, and the last update to Zlib was April 2013, so we're indeed talking about 15 years.  What are the changes?  Well, you can see the [CHANGELOG.TXT](http://www.zlib.net/ChangeLog.txt).  The first patch after the version you used was in 2002, which lists:
- ZFREE was repeated on same allocation on some error conditions.  This creates a security problem described in [http://www.zlib.org/advisory-2002-03-11.txt](http://www.zlib.org/advisory-2002-03-11.txt)
- Returned incorrect error (Z_MEM_ERROR) on some invalid data
- Avoid accesses before window for invalid distances with inflate window less than 32K.
- force windowBits > 8 to avoid a bug in the encoder for a window size of 256 bytes. (A complete fix will be available in 1.1.5).

It goes on from there; though a large part of what they've been doing is in makefiles, examples, just basic cross-compiler compatibility of the source.

Your point about picking an anchor and sticking to it is valid.  But I think that's more about when some library grabs a heinous dependency on something else, not a simple 1K per year set of patches and improvements to the same basic stuff.  If you're going to freeze time, you also become the maintainer of that frozen time.  :-/  This doesn't seem like the right place to be doing that.

However, the granularity of inclusion in C is function-level.  I didn't go line-by-line to see what pieces of the code could be #ifdef'd out as stuff Rebol does not use...or some optimization trading performance for code size that isn't a useful balance.  It just seemed like something that needed to be done so I subsetted it to what Rebol would use.  It fixed a bug that I traced as not being related at all to the code path in modern Zlib so it seemed better to start from modern Zlib when fixing it, but using the modern Zlib fixed it... so... I thought it the best avenue.

If 18.5K seems high, then we can probably knock a K off here and there, if that's a priority.  I still advise doing that as part of the subsetting script, though.  It does start to get a little tricky but Zlib isn't updated _that_ quickly, and now that Rebol is open source whoever does a "new Zlib" update with the scripts could go look closely to make sure any exclusions were still correct.

But for 18.5K... is it worth it?  I think that the real time to go in and micro-optimize the size is when an application has called for it.  It's not the absolute number of bytes in a lib, it's the _dependencies_ to really be watching out for.

BrianHawley:
Do we need to have a build option to link to a system-provided zlib? It would shrink things on Android, for instance, or other platforms where zlib is guaranteed to be included with the OS.

hostilefork:
@BrianHawley That was part of my idea.  Of course, the option then ties you to interface changes on whatever the environment picks and you wind up in your own version of "DLL hell".  All things being equal, it's probably best to pack your own copy unless you've got a really specific circumstance.

I was just making the point that it isn't really about micro-optimizing _byte size_...it's about _dependency graph size_.  Failure to do so is why you can't install a tiny utility on Linux without pulling down megabytes.  It's not because of 18K of patches and algorithms, it's because someone dragged in one routine from a library that called another...next thing you know you've got megabytes for something simple.

So by default the binaries should dynamically link as little as possible...but using the OS versions should be an option.  The OpenSSL code has some inline assembly for the crypto, and I don't think introducing a requirement for an inline assembler for Rebol is a good idea.  But if the makefile had an option to link to it dynamically if the host had compiled for that... why not?  It's just part of being up to date.

Trying to think of slogans: _"Count dependencies, not bytes.  Then count the bytes.  And realize that counting the dependencies is counting the bytes, and the troubles."_  Adjust motto from there.

BrianHawley:
In the case of Android, the NDK includes a static library `libz.a` which links to the system-provided zlib dynamic library, which is mandated to be included in Android. Any NDK app can link to that static lib and work on any Android version. It's a bit of a different situation when you have known libs with guarantees that they'll be there and linkable. We shouldn't include dependencies on non-mandatory libs, but the mandatory libs should be OK.

hostilefork:
Because @Cyphre wrote the missing codec tests, he tripped the PNG bug that motivated me to do this in the first place.  His change is one line:  

https://github.com/cyphre/r3/commit/a5fea464679656e1e0c005379b8c866b676077fb

My response of building the packager from the current repository was based on looking at the crash line and finding it didn't exist in modern Zlib for comparison.  I didn't feel like debugging 15 year old code.

Accepting Cyphre's change would remove the urgency on this one for me.  Yet I'll again point to the [CHANGELOG.TXT](http://www.zlib.net/ChangeLog.txt).  Some changes make Zlib build better on newer 64-bit platforms or adapt to older platforms better... they still build in Turbo C and OS/2 and try to keep that stuff working, so they're philosophically aligned with the Rebol mission.  Some speed it up...some fix bugs.  Some make it compile better with C++ compilers (hint, hint).  A lot deals with how error conditions are handled; and while errors are by definition rare, this is where security holes and debugging nightmares come from; it's never a bad idea to tighten the screws.

Yet many fix documentation or examples, which isn't relevant here.  I just did a quick glance over to pull things that look non-trivial:
- ZFREE was repeated on same allocation on some error conditions.  This creates a security problem described in [http://www.zlib.org/advisory-2002-03-11.txt](http://www.zlib.org/advisory-2002-03-11.txt):
- Returned incorrect error (Z_MEM_ERROR) on some invalid data
- Avoid accesses before window for invalid distances with inflate window less than 32K.
- force windowBits > 8 to avoid a bug in the encoder for a window size of 256 bytes. (A complete fix will be available in 1.1.5).
- Use malloc() instead of calloc() in zutil.c if int big enough
- Fix static const's in deflate.c, gzio.c, and zutil.[ch]
- Fix uncompress() to return Z_DATA_ERROR on truncated input
- Fix bug in gzread() on concatenated file with 0x1f at end of buffer
  and next buffer doesn't start with 0x8b
- Fix bug in gzopen() on single-byte file 0x1f
- Check for some malloc() failures in gzio.c
- New and improved crc32() _(About 50% faster, thanks to suggestions from Rodney Brown)_
- Fix memory leak in deflateInit2()
- New and improved inflate code _(About 20% faster, Does not allocate 32K window unless and until needed, Automatically detects and decompresses gzip streams, Raw inflate no longer needs an extra dummy byte at end, Added inflateBack functions using a callback interface--even faster than inflate, useful for file utilities (gzip, zip), Added inflateCopy() function to record state for random access on externally generated deflate streams (e.g. in gzip files))_
- Add Z_RLE strategy for run-length encoding [Truta]
- Correct Z_BLOCK to not return on first inflate call if no wrap
- Fix strm->data_type on inflate() return to correctly indicate EOB
- Fix a big fat bug in inftrees.c that prevented decoding valid dynamic blocks with only literals and no distance codes -- Thanks to "Hot Emu" for the bug report and sample file
- Fix trees.c to update strm->data_type (no one ever noticed!)
- Fix bug in error case in inflate.c, infback.c, and infback9.c [Brown]
- Provide two versions of all functions that use the z_off_t type for binary compatibility -- a normal version and a 64-bit offset version, per the Large File Support Extension when _LARGEFILE64_SOURCE is defined; use the 64-bit versions by default when _FILE_OFFSET_BITS is defined to be 64
- Fix EOF detection in transparent gzread() [Maier]
- Fix bug in adler32_combine() [Zhu]
- Catch missing-end-of-block-code error in all inflates and in puff _(Assures that random input to inflate eventually results in an error)_
- Correct error return bug in gz_open() by setting state [Brown]
- Fix big fat stinking bug in gzseek() on uncompressed files
- Fix memory leak on error in gz_open()
- Fix bug in gzeof() to take into account unused input data
- Fix bug in gzread.c when end-of-file is reached
- Remove code from inflate.c and infback.c that is impossible to execute
- Change gzgetc() to a macro for speed (~40% speedup in testing)
- Change gzread() and related to ignore junk after gzip streams
- Fix bug in gzgets() for a concatenated empty gzip stream
- Fix bug in gzclose() when gzwrite() runs out of memory

hostilefork:
Subsetting script and update incorporated into [Coherence 1.5](https://github.com/metaeducation/ren-c/pull/12); closing this PR.

PR #156, opened 2013-09-30
Free intermediate buffers used by do-codec (CC #2068)

hostilefork:
Leak from [CC #2068](http://curecode.org/rebol3/ticket.rsp?id=2068).  Found with valgrind but it would have been obvious even without.  Valgrind confirms no leaks afterward.

These two calls to Free_Mem seem to resolve the issue for now.  But it looks like the codec extensibility model is skeletal and ad-hoc at the moment.  See also this question:

http://stackoverflow.com/questions/14365034/how-do-you-write-a-codec-for-rebol-3

A general review of codecs that will work for letting people supply codecs written as Rebol code is needed.  That as well as native JPEG and GIF encoding (as opposed to just decoding)...choosing which codecs to include in minimal builds... among many other considerations.

### DIFF
--- a/src/core/n-system.c
+++ b/src/core/n-system.c
@@ -450,14 +450,18 @@ err:
 		memcpy(BIN_HEAD(ser), codi.data, codi.len);
 		Set_Binary(D_RET, ser);
 		if (result != CODI_BINARY) VAL_SET(D_RET, REB_STRING);
-		// Free the binary!
+		
+		// See notice in reb-codec.h on reb_codec_image 
+		Free_Mem(codi.data, codi.len);
 		break;
 
 	case CODI_IMAGE:
 		ser = Make_Image(codi.w, codi.h, TRUE); // Puts it into RETURN stack position
 		memcpy(IMG_DATA(ser), codi.bits, codi.w * codi.h * 4);
 		SET_IMAGE(D_RET, ser);
-		// Free the image!
+
+		// See notice in reb-codec.h on reb_codec_image 
+		Free_Mem(codi.bits, codi.w * codi.h * 4);
 		break;
 
 	case CODI_BLOCK:
--- a/src/include/reb-codec.h
+++ b/src/include/reb-codec.h
@@ -29,6 +29,17 @@
 #define CODI_DEFINED
 
 // Codec image interface:
+//
+// If your codec routine returns CODI_IMAGE, it is expected that the
+// ->bits field contains a block of memory allocated with Make_Mem
+// of size (->w * ->h * 4).  This will be freed by the
+// REBNATIVE(do_codec) in n-system.c
+//
+// If your codec routine returns CODI_BINARY or CODI_TEXT, it is
+// expected that the ->data field contains a block of memory
+// allocated with Make_Mem of size ->len.  This will be freed by
+// the REBNATIVE(do_codec) in n-system.c
+//
 typedef struct reb_codec_image {
 	int action;
 	int w;


ladislav:
I think that it would be ideal to have some tests in the test framework for this, could you write some?

hostilefork:
The trick with writing tests for memory leaks is that if you really want to get to the point of crashing, it will probably have to get pretty far into making your system nearly die from VM exhaustion.  Not only does it take time, it makes the machine flaky and unusable.

So the better thing to do is to check for memory leaks with Valgrind as part of the automated testing.  This is tricky as there are crashing tests which would destroy the evidence of leaks in tests prior to the crash.  And one probably doesn't want to start up a new interpreter for each line-item test.

What I'd suggest is that the automated tests probably should let you mark a test with a tag indicating it is expected to crash.  Then, the higher-level script in charge of the test would be sure to shut down the interpreter before that point, and check to make sure there weren't any leaks.  Then it would start back up to run the suspected crashing test.

ladislav:
Merged to community.

cyphre:
There is a regression caused by this commit. Following code now crashes R3:

```
do-codec system/codecs/text/entry 'decode #{}
```

This is caused by the fact the `codi.data` buffer is in the `TEXT` codec case freed while it shouldn't as it is actually the binary! input argument of `do-codec` call. See the line:

```
Free_Mem(codi.data, codi.len);
```

@carls I'd suggest to accept commits only covered with proper tests otherwise the master will be soon spoiled by these small but annoying regresssions. It looks from the previous discussion Ladislav smelt something fishy is there.

earl:
Thanks a lot for the testing and the report, @cyphre. I added a test for the reported regression in rebolsource/rebol-test@10c0160c21951ead624df3e4af381d52388206d4.

I think we shouldn't exaggerate this problem. Bugs happen, let's just fix them. And let's keep the test suite growing.

cyphre:
@earl I agree bugs happen but having at least some basic test coverage for specific fix would avoid these kind of regressions. I'm talking this case just as an example, this is nothing personal against Brian at all. IMO it would be usefull policy that every contributor should provide at least basic testing otherwise other people need to spend time dealing with the regressions anyway.
(BTW Nowadays many other dev environments won't even let you use your code/changes without coverage tests)

hostilefork:
@cyphre If you looked and understood what's going on well enough to fix it, then perhaps a pull request would be of more utility than something more enigmatic.  Five months later I am working on something else.

Further: To write tests, one generally needs a spec.  And your example case: `do-codec system/codecs/text/entry 'decode #{}` is literally greek to me (er, okay... I guess that would be "figuratively").  Point is: I don't know what that means _at all_.

What happened here is someone was engaging the official bug database and asking on SO questions and chat about how to forge ahead with a Raspberry pi project.  It was a blocking bug for him; and not addressing it can risk someone not using Rebol at all.  I looked at it enough to make it work for the case I was told about; without undertaking a total design review of the codec system which (as I mentioned) has not been documented.

Without a spec I cannot be reasonably expected, as an unpaid contributor to _design_ a test that would even _know_ that `do-codec system/codecs/text/entry 'decode #{}` was a pattern in existence.  So nothing personal to you either... but there were _5 months_ of time to review this.  Even if you paid me, I can't design tests for things that aren't defined...unless I'm paid for the definition too!

So I'd ask you to shift perspective to see another viewpoint:
- unblocking an _active_ user request _in public forums_ while breaking no existing test
- not having a spec to define new tests from
- no feedback or interaction on public code, with 5 months of review time
- paid $0 for the privilege of investigating a problem that's not mine... _which no one else was stepping up to address or respond to_

I'd summarize by saying my "lessons we should carry away" from the instance is different from yours.  If we choose to be more careful, then do we take 50 months instead of 5 to resolve an issue and lose users in that time?  @earl is on point that the real answer is more tests, coupled with more forward progress.  We need more public concern and engagement from everyone, on a daily basis, which is more significant than the impacts of the regression.

Though I would of course welcome a fix if you have one in mind.  I'd have to go back and re-acclimate to the issues, and it seems like it's fresher in your mind.

ladislav:
I need to throw my two cents to the issue. In general, many PRs _are_ accompanied with tests that are added to the test suite frequently _before_ the issue is being worked on. Therefore, the PRs based on such tests are more reliable than PRs not having any tests associated. It is good that this PR exists, but, by my own fault, it has been merged to Community prematurely (before any test existed and has been ran) and I realize that I should not have done that. I am sorry for the trouble it caused to you, Cyphre, and I promise to be more careful next time. For contrast, consider #153 well tested (equipped with several tests in the test suite) and yet, Carl hesitates to merge it to the master branch as compared to several PRs like this one unequipped with any tests at all. This indicates there is a merging policy issue to improve on.

cyphre:
@hostilefork I pushed the pull request you asked for here https://github.com/rebol/rebol/pull/194
If you look at the PR you'll find also included Gist with all the possible tests for the function that was changed. If you run it with previous build you'll easily dicover crash or bug in any current codec functionality.
Don't you think such PR is easier for Carl to merge and be more sure it won't most probably break anything?
If contributors accept such policy it makes life easier not only to Carl when merging but also to people that are using the merges thinking they are at least 'somehow' tested.
In other words, if the contributor write a coverage test it also guarantee he spent some time looking into the problem and not only quick-hacked something in.

BrianHawley:
@cyphre: "Nowadays many other dev environments won't even let you use your code/changes without coverage tests"

And that is the problem. While it would be a good policy to only merge if the tests pass, _it's clearly not our policy here_. Our test suite is not in this repo, it's in another repo controlled by someone else. We can't make changes that might require changing tests (such as everything I've been working on lately, we have a backlog), because the tests are not kept in sync with the changes. The way things are now, it's just not a practical policy to enforce, and we have no right to berate others for not doing what we're clearly not doing either.

Some steps we can take to fix the problem:
- Move the test suite into this repo.
- Add support for marking tests as "pending", then mark every test that doesn't pass in master as such. There should never be a test that fails.
- Make the test suite able to refer to fixtures (external files), since many of the tests that aren't in that suite at the moment require such files.
- Change the test suite so it can be run by a continuous integration server. There are free ones.
- Set up the repo with said CI server, to compile and run the test suite with every PR, and mark it as failing if it doesn't pass.

Once we do all that, we will have the right to say that including tests is our policy. And we need to do it soon, because the basic-semantics fixes are coming.

hostilefork:
@ladislav 

Point of clarification: I actually did not understand your request as a desire to add tests which would be _able to deduce what this change might disrupt_.  I thought you were asking for a test that would _verify that the prior existing condition of a memory leak did not reappear_.  That codec tests did not exist at all was not known to me.

@cyphre 

Thanks for submitting the update, and expanding upon the comments I added to the code.  I want to emphasize that my disagreement here is of course not with the need for more testing.  I'm rather stating that raising the bar for _submitting_ a pull request--to a point where the submitter must design test suites for previously untested subsystems--simply isn't going to fly on an open source project with contributions from a general user community.

If the integrators wish to hold off on pulling in a patch, then that's one thing.  But I definitely don't want to discourage the submission of changes such as this one, from others (or myself) who make what looks like a reasonable change that fixes a blocking problem.  As I emphasized, it was already something I was doing as a favor to help someone else move along.  When it took care of their problem, and passed the test suite _that was enough for me_...and so it was submitted to be looked at.

It's fair to say a pull request will not be reviewed until it passes the existing tests, and CI will help with that (also possibly in testing platforms the user might not have had available).  But I maintain the process problem here _isn't the pull request being made_.  It's actually really three other points:
- Initial code written before spec / document / tests were established
- Absence of development community members on GitHub to review and comment on pull requests _(...until after the fact!)_.  It's mostly been falling all to just Ladislav and @earl, and everyone (myself included) should be more involved in reading the diffs... making annotations on the lines if anything jumps out at them, etc.
- I'll agree it was integrated prematurely, now that I know there were _no_ codec tests

So that's my take on it.

@BrianH

Agreement on pretty much all points.  I was in fact just going to suggest something like your "pending", due to feeling that the need to diff the logs to tell there weren't any regressions was excessive.  While pending should be called out separately from a test where failure is the expected/correct result, the test should still be run... and a pending test that passes should count as an error.

If Rebol and Red are to share tests, then the shared tests should probably stay in a separate repository.  Submodules can give the effect of it feeling like the directory is part of the master despite the sharing.  Even writing just one test that is successfully shared between Rebol and Red, testing that **2 = 1 + 1**, would be a step in the direction of beginning to identify common semantics.

Note: Automated testing should also look for bootstrap breaks and warn about those.  Both the current Rebol and a reasonably older Rebol should be able to do **make make** and **make prep**.

PR #158, opened 2013-10-01
Fix notorious TCP bug (EVM_DEVICE events allowing ports to GC)

hostilefork:
Each Rebol value is implemented internally as C structures.  For the garbage collector to know which Rebol series (or objects) are being referenced by a value, this C structure must be scanned on a per-type basis.  For simple terminal types like an INTEGER! or a CHAR!, there are no embedded types.  Clearly BLOCK! holds onto the string series and other values it holds... but there are other instances, which can be subtle (a FUNCTION! holds its spec-block and body-block, for instance... as well as an internal optimized list of arguments generated from the spec.)

The [`Mark_Series()`](https://github.com/rebol/rebol/blob/4d9840f8e661e7067472e828ee6ad8dd3008b247/src/core/m-gc.c#L148) routine is responsible for sniffing the type of a REBSER, and take care of the type-specific behaviors for its elements.  In two cases it would call type-specific functions in the switch statement: for GOB! and EVENT! objects.  The GOB! function was delegated to make as many calls as it wanted to MARK_SERIES or CHECK_SERIES to handle its internal structures, while the EVENT! function was merely to return a single REBSER that it might be referencing.

This pull request does the following:
- Adds handling in the EVENT! GC marking handler to account for the case of being an EVM_DEVICE event type, in which the PORT! object being held onto is not in the VAL_EVENT_SER of the REBEVT structure, rather held indirectly by the REBREQ ->req field of the event.  Failure to account for this before was leading to crashes resulting from ports being improperly garbage collected; the "notorious TCP bug".
- While looking at the REBREQ field, I noticed a ->next field for a linked list of more REBREQs.  As the previous problem had come from a port being GC'd out from under a REBREQ, I was concerned that these other requests might be holding ports that were similarly unaccounted for.  @earl explained that if a request had made it into an event, then it had been processed and the ->next field would be NULL.  The ->next field was only used by REBDEV devices when holding requests in a pending queue.  In a "better safe than sorry" philosophy I added a `Mark_Pending_Requests()` function to walk the pending queue for the devices during a Recycle() and mark the ports they hold as used... it does pick up some ports, so this could well be an issue.  However the current layering doesn't support a way to cleanly get at devices pending queues from the GC so the "unclean" code is under **#if NEED_TO_MARK_PENDING_REQUESTS**
- Stylizes the event GC marking routine with interface `Mark_Event` to match the working of `Mark_Gob`, for consistency.  This required removing the "static" keyword from Mark_Series which previously limited it to being called only from m-gc.c.  Potential for the routine being called incorrectly is mitigated by making sure a GC is in progress when it is called.  For consistency and ease of understanding of the code, it's reorganized so that `Mark_Gob` lives in the t-gob.c file (as the event! GC marking routine had been in t-event.c).  For code comprehensibility, moving the logic of understanding the internal structure of the types into their appropriate t-*.c files makes more sense than having that structure exported to the garbage collector.  This is a step toward a modular approach for the per-type callout routines.

### DIFF
--- a/src/core/m-gc.c
+++ b/src/core/m-gc.c
@@ -79,6 +79,9 @@
 
 #include "sys-core.h"
 
+// Track whether we are currently running the garbage collector
+REBOOL In_Recycle = FALSE;
+
 //-- For Serious Debugging:
 #ifdef WATCH_GC_VALUE
 REBSER *Watcher = 0;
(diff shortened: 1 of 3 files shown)


hostilefork:
@Ladislav The check wasn't perfect but was easier than adding a new variable, but if it's a new variable you want, you've got it.  `In_Recycle`, updated.

The idea that an Event has a single series to watch out for with a mark was called into question by the apparent "list of requests" that was held by an event, due to the request's ->next.  So my original motivation for changing the function from GC_Event returning a series to Mark_Event was based on making it possible for the type to mark multiple things (as GOB does).

The better modularity would come from having _"Point out the parts of a structure that are holding elements that need to be GC'd"_ in the file related to that structure.  If you believe otherwise, you would think that `m-gc.c` should contain includes so it can evaluate expressions like `(VAL_EVENT_TYPE(value) == EVT_DROP_FILE && GET_FLAG(VAL_EVENT_FLAGS(value), EVF_COPIED))`?

So by removing the static annotation on the `Mark_Series` routine, the logic for how GOB! points out the series it holds onto was unchained from being in `m-gc.c`.   Having the garbage collector knowing too much about the internal structure of types doesn't strike the right modularity, and the other types should be handling it closer to the rest of the source related to the internal structure of that type too.  This was just a start, since it was already a larger and separate function from Mark_Series.

ladislav:
For an alternative minimized version of this PR see #161 .

earl:
I prefer the variant in #161 over this one. Even from a modularity perspective, in this particular case I prefer keeping the GC code together over keeping the type internals together. That's the age old type extensibility over operation extensibility debate. Again, in this case I prefer the GC code kept together. If we think the GC too invasively inspects the type structure, I'd rather provider better type-specific wrappers for whatever information the GC needs.

However, this pull request raises the important "`NEED_TO_MARK_PENDING_REQUESTS"` question. We must not forget about this issue, so I suggest we record it in CC for a start. We'll either have to implement a proper fix for that (and fix the layering first), or at least leave a documenting comment explaining why pending requests need not be marked.

carls:
Accepted #161 but with the caveat that we need to resolve Brian's question.

earl:
Recorded the "pending requests issue" as [CureCode issue 2106](http://issue.cc/r3/2106).

PR #161, opened 2013-10-02
This is an alternative to HostileFork's "notorious TCP bug fix"

ladislav:
(EVM_DEVICE events allowing ports to be collected) minimized. It is meant as an alternative to the pull request #158 .
Tested under Linux (0.4.4) with no regressions.

Differences:
- keeps marking code in the garbage collector sources
- no need for special tests whether parts of the GC are called from inappropriate places
- not making changes unrelated to the problem
- traverses the list of requests

### DIFF
--- a/src/core/m-gc.c
+++ b/src/core/m-gc.c
@@ -22,7 +22,7 @@
 **  Module:  m-gc.c
 **  Summary: main memory garbage collection
 **  Section: memory
-**  Author:  Carl Sassenrath
+**  Author:  Carl Sassenrath, Ladislav Mecir, HostileFork
 **  Notes:
 **    WARNING WARNING WARNING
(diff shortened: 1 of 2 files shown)


hostilefork:
The issue of whether or not pending requests may need to have their ports held onto in the GC is an important one, and this sweeps that under the rug completely, rather than leaving literate _working code_ in to be studied and kept in mind.

ladislav:
Merged to community.

carls:
Change can be boiled down to moving the event garbage sweep into the GC module as Mark_Event, and most importantly. making it sensitive to device request field linkage. This change is simple, but as hostilefork points out, it may be incomplete, so the question should be resolved so we can decide if anything else from PR #158 needs to be added.

PR #164, opened 2013-10-03
Meanings of TRUE and FALSE w.r.t. INTEGER! for TO vs. MAKE (CC #2055)

hostilefork:
Very simple change implementing the idea ([CC #2055](http://issue.cc/r3/2055)) that things like TO LOGIC! 0 are TRUE, while things like MAKE LOGIC! 0 remains false.  The converse is that TO INTEGER! TRUE or TO INTEGER! FALSE return errors, while MAKE INTEGER! TRUE returns 1 and MAKE INTEGER! FALSE returns 0.

Does not remove the `true?` or `found?` mezzanines, as consensus on that did not seem to have been reached, and @earl believes that `true? x` can read better than `to logic! x`.  Separate ticket(s) should be opened for those issues.

Three core-tests affected:
- `[false = to logic! 0]` [link](https://github.com/rebolsource/rebol-test/blob/master/core-tests.r#L1793)
- `[0 = to integer! false]` [link](https://github.com/rebolsource/rebol-test/blob/master/core-tests.r#L1692)
- `[1 = to integer! true]` [link](https://github.com/rebolsource/rebol-test/blob/master/core-tests.r#L1693)

### DIFF
--- a/src/core/t-integer.c
+++ b/src/core/t-integer.c
@@ -202,53 +202,61 @@
 
 	case A_MAKE:
 	case A_TO:
-		val = D_ARG(2);
-		if (IS_DECIMAL(val) || IS_PERCENT(val)) {
-			if (VAL_DECIMAL(val) < MIN_D64 || VAL_DECIMAL(val) >= MAX_D64)
-				Trap0(RE_OVERFLOW);
-			num = (REBI64)VAL_DECIMAL(val);
-		}
-		else if (IS_INTEGER(val))
-			num = VAL_INT64(val);
-		else if (IS_MONEY(val))
-			num = deci_to_int(VAL_DECI(val));
-		else if (IS_ISSUE(val)) {
-			REBYTE *bp;
-			REBCNT len;
-			bp = Get_Word_Name(val);
-			len = strlen(bp);
-			n = MIN(MAX_HEX_LEN, len);
-			if (Scan_Hex(bp, &num, n, n) == 0) goto is_bad;
-		}
-		else if (IS_BINARY(val)) { // must be before STRING!
-			REBYTE	*bp;
-			n = VAL_LEN(val);
-			if (n > sizeof(REBI64)) n = sizeof(REBI64);
-			num = 0;
-			for (bp = VAL_BIN_DATA(val); n; n--, bp++)
-				num = (num << 8) | *bp;
-		}
-		else if (ANY_STR(val)) {
-			REBYTE *bp;
-			REBCNT len;
-			bp = Qualify_String(val, MAX_INT_LEN, &len, FALSE);
-			if (memchr(bp, '.', len)) {
-				if (Scan_Decimal(bp, len, DS_RETURN, TRUE)) {
-					num = (REBINT)VAL_DECIMAL(DS_RETURN);
-					break;
-				}
+		{
+			REBOOL make = (action == A_MAKE);
+			val = D_ARG(2);
+			if (IS_DECIMAL(val) || IS_PERCENT(val)) {
+				if (VAL_DECIMAL(val) < MIN_D64 || VAL_DECIMAL(val) >= MAX_D64)
+					Trap0(RE_OVERFLOW);
+				num = (REBI64)VAL_DECIMAL(val);
 			}
-			if (Scan_Integer(bp, len, DS_RETURN))
-				return R_RET;
-			goto is_bad;
+			else if (IS_INTEGER(val))
+				num = VAL_INT64(val);
+			else if (IS_MONEY(val))
+				num = deci_to_int(VAL_DECI(val));
+			else if (IS_ISSUE(val)) {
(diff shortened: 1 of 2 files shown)


carls:
The use of the scoping block and extra indentation makes the git diff fail. I've tried to avoid statement-less blocks because various older C compilers will choke on them. I suspect this code will probably not compile on the Sun, HPUX, and various BSD systems I have here. So, the benefit of the cleaner scoping has a cost. (And that pattern is not used in Rebol core anywhere.)

cyphre:
Just spent 2 hours finding regression in R3 script that was caused by the to logic! 0 change. I think we should be more careful when applying such changes. I don't see any benefit of the current functionality:

```
>> to logic! 0
== true

>> make logic! 0
== false
```

IMHO it makes the MAKE vs TO  area even more messy.

For example now this code:

```
constructed?: to logic! (d and 32)
```

will look like:

```
constructed?: make logic! (d and 32)
```

Try to explain that difference to some Rebol newbie...

hostilefork:
@cyphre Sorry it was an inconvenience.  :-(  Though for better or worse, Rebol's definition of logic is that 0 is true; and users are asked to buy into that.  It is unfortunate the change will break some code...but it is simply more consistent.

However: the rationale for not changing the behavior for MAKE wasn't to encourage people to convert existing TO LOGIC! instances to use MAKE!  It was rather the idea of not changing more things than were necessary in that moment.  Note that MAKE doesn't pretend to be a "conversion" of its input but rather a "construction" using its input. _(Compare `make block! 10` with `to block! 10`)_.  For better or worse, MAKE has no preconceptions hanging over its head...so it has the liberty to be as wacky as it wants to be.  Yet TO and `.to_s`or other conversions are going to be things that people come in with expectations about.

So for your change, I'd certainly avoid switching to MAKE LOGIC!...and instead being explicit about the comparison against zero:

```
 constructed?: zero? (d and 32)
```

That's pretty easy to explain.  At least relative to explaining why `if to logic! 0 [print "true"]` won't print true but `if 0 [print "true"]` will.  :-/

**But one thing is certain:** there should be a system of notifications of what patterns to look for when invariants change, and how to respond.  I know from trying to do the [Rebol2 to Rebol3 port of Red](https://github.com/red/red/pull/421) that changes which don't error out are much harder to find than the ones that do.  For instance, in Rebol3 SHIFT changed from right shift by default to left shift.  :-/

So if things are going to change--_and they need to_--the real question is how to get notices out to people.  Given that people are using disparate tools to gather information it seems the best idea is to create some kind of feed that can be echoed multiple places.  Another issue is bringing people together in a common place where the discussions involve _everyone_.  That's not happening right now, and it needs to.  Even if people don't have time for the fray of the discussions in CureCode, it's imperative that stakeholders find the time to review things once a semantic change has come to the point of being a pull request.

_(In this case, there was a full 5 month review period for the pull request... not counting the time for the discussion prior.)_

Another thing is finding ways to move in phases.  Random example: this would be an easier change to integrate if TO got a refinement called /LOGIC0 (or something) and all TO would throw an error if they got in a LOGIC! and a value of 0 without the refinement.  This could be temporary and eventually phased out.  Perhaps not the best suggestion but just presenting a line of thought about how to make invariant changes create only places to update in code, vs. hard-to-track-down-bugs.

cyphre:
When talking about TO vs. MAKE I'd personally get rid of TO completely in R3 and use only MAKE with optional /ALLOC refinement for series or something (to get rid of all the differences between these two commands) but that would be rather more radical change :-)

I understad your point but I don't like if to logic! 0 [print "true"] will print true now while if make logic! 0 [print "true!"] not. IMO There is no logical reason why should be MAKE more or less relaxed and we'll be bitten by this everytime when not looking into some "reference table" with tens or hundreds possible conversion/creation results.
This area seems to me confusing but I'm afraid there is probably not much to do with it so we have to live with that.

hostilefork:
@cyphre I think there are probably not a lot of instances of **make logic!** out there and you have brought up a good point that this change should not be reacted to by introducing more.  It does indeed expand the space just as **make block! 10** vs. **to block! 10** does, and that prior duality was some of the "justification" for why it was "no worse than things were already".  Not the best argument, perhaps.

Radical suggestions don't scare me if they make sense.  Having TO and MAKE does mean it's _(2 \* [input type count] \* [output type count])_.  It's hard in the first place just to have the cross product, and even harder to have it multiplied by 2.  One shouldn't be afraid to reason through it and speak up; it can at least suss out documentation of why things have to be the way they are.

The CureCode issue is still there and if you have design points to raise that's probably the place to do it for now: [CC #2055](http://issue.cc/r3/2055) -- @DocKimbel has expressed specific interest in knowing what the "brain trust" has to say about TO and MAKE, because it is a large number of cross products to manage and there's not a lot of clarity surrounding it.  More clarity would be good.

PR #166, opened 2013-10-03
Default splice matching block types else use explicit /SPLICE (CC #2057)

hostilefork:
This implements the nicer invariant that splicing of any-block types for operations such as INSERT, APPEND, and CHANGE only happens if the block types match ([CC #2057](http://issue.cc/r3/2057)).  Otherwise you have to ask for it explicitly:

```
>> append [a b c] [d e f]
== [a b c d e f]

>> append [a b c] 'd/e/f
== [a b c d/e/f]

>> append/splice [a b c] 'd/e/f
== [a b c d e f]
```

For the moment, /splice overrides /only and does not signal an error, if for some reason you use both.  I'd probably prefer an error myself, but @BrianH wants to avoid conflicting-refinement-errors:

```
>> append/only/splice [a b c] 'd/e/f
== [a b c d e f]
```

The PARSE behavior for CHANGE and INSERT required modification as well to support this.  If you have `x: [1 2 "foo" "baz" "bar" 3 4]` then this is how it goes for some various cases:

```
>> parse x [some [to string! change [some string!] [mumble]]] probe x
[1 2 mumble 3 4] ;-- as before

>> parse x [some [to string! change [some string!] only [mumble]]] probe x
[1 2 [mumble] 3 4] ;-- as before

>> parse x [some [to string! change [some string!] (mumble)]] probe x
[1 2 (mumble) 3 4] ;-- new behavior, dropped parens before by default

>> parse x [some [to string! change [some string!] splice (mumble)]] print x
[1 2 mumble 3 4] ;-- new behavior, splice joins only as parse keyword
```

If your x was instead in parens, then this would be different, e.g. with `x: quote (1 2 "foo" "baz" "bar" 3 4)` for these cases:

```
>> parse x [some [to string! change [some string!] (mumble)]] probe x
(1 2 mumble 3 4)

>> parse x [some [to string! change [some string!] [mumble]]] probe x
(1 2 [mumble] 3 4)
```

### DIFF
--- a/src/boot/actions.r
+++ b/src/boot/actions.r
@@ -269,23 +269,25 @@ take: action [
 ]
 
 insert: action [
-	{Inserts element(s); for series, returns just past the insert.}
+	{Inserts item, or contents if block type match; returns insert tail.}
 	series [series! port! map! gob! object! bitset! port!] {At position (modified)}
 	value [any-type!] {The value to insert}
 	/part {Limits to a given length or position}
 	length [number! series! pair!]
 	/only {Only insert a block as a single value (not the contents of the block)}
+	/splice {Force splicing semantics even for mismatched any-block! types}
 	/dup {Duplicates the insert a specified number of times}
 	count [number! pair!]
 ]
 
 append: action [
-	{Inserts element(s) at tail; for series, returns head.}
+	{Inserts item at tail, or contents if block type match; returns head.}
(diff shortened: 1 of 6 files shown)


ladislav:
-1

See the discussion at [CC#2057](http://issue.cc/r3/2057) why this PR is incompatible with Rebol design.

hostilefork:
Whatever.

PR #167, opened 2013-10-03
Fix CATCH/quit interaction with TRY (cc#851)

earl:
As reported in [CC#851](http://issue.cc/r3/851), currently, a `catch/quit` incorrectly disables other (outer) error handlers, such as `try` or `attempt`.

The underlying cause is the `catch/quit` implementation incorrectly overwriting the error handler stack. See the commit message for more details on the fix.

This fixes [CureCode issue #851](http://issue.cc/r3/851).

### DIFF
--- a/src/core/c-do.c
+++ b/src/core/c-do.c
@@ -1590,6 +1590,9 @@ eval_func2:
 **
 */	REBOOL Try_Block_Halt(REBSER *block, REBCNT index)
 /*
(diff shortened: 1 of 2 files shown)


ladislav:
Fixes also [CC#1896](http://issue.cc/r3/1896).

earl:
Tested on Linux/x86: no regressions, 3 progressions.

ladislav:
Merged to community.

PR #169, opened 2013-10-04
PARSE evals functions and closures to produce rules (CC #2075)

hostilefork:
An implementation attempt at [CC #2075](http://curecode.org/rebol3/ticket.rsp?id=2075).  This could well not be quite right, but that's why I'm submitting it for review.  It appears to basically work, but appearances can be deceiving, so...

I noticed that in order to use Do_Next, the point of invocation needed to pass in a series and an index.  Parse didn't have the actual REBSER on hand once you got into the guts of it... it was passed the VAL_BLK_DATA and then iterated to the end of it.  So most of the change was just adding it to the function interfaces as `rules_block`.  To avoid confusion I then renamed `rules` to just `rule`.

I'm a little confused about stack handling and lifetimes of the generated results by Do_Next, and how safe this is w.r.t. garbage collection and volatility.  I also don't know if the existing routines which seemed to put things on the top of the stack and then read them with DS_TOP, without ever doing a DS_POP, are actually leaking things onto the stack memory.

So some eyes on this will definitely help.

It's a little awkward to deal with getting passed a rule position that is past the evaluation position... so `Eval_Parse_Expression_Volatile` is weird.  Did it this way to keep from changing too much code right now, but this could be made more sensible in a code reorg.

### DIFF
--- a/src/core/u-parse.c
+++ b/src/core/u-parse.c
@@ -82,7 +82,7 @@ void Print_Parse_Index(REBCNT type, REBVAL *rules, REBSER *series, REBCNT index)
 
 /***********************************************************************
 **
-*/	static REBCNT Parse_Series(REBVAL *val, REBVAL *rules, REBCNT flags, REBCNT depth)
+*/	static REBCNT Parse_Series(REBVAL *val, REBVAL *rules_block, REBCNT flags, REBCNT depth)
 /*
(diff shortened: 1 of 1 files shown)


BrianHawley:
This requires more discussion before we go forward. See the ticket for that discussion.

PR #170, opened 2013-10-04
IF, EITHER, and UNLESS accept all values in then/else slots (CC #2063)

hostilefork:
Implementation of [CC #2063](http://curecode.org/rebol3/ticket.rsp?id=2063) which makes these natives allow any value in their then/else slots.  Blocks are evaluated as before.

Needed to add R_ARG4, but wouldn't have to in the case that `if/else` were abandoned (which it should be!  no one wants to get Rebol code with `if/else` in it...)

### DIFF
--- a/src/core/n-control.c
+++ b/src/core/n-control.c
@@ -642,7 +642,13 @@ got_err:
 		DO_BLK(D_ARG(argnum));
 		return R_TOS1;
 	} {
-		return argnum == 2 ? R_ARG2 : R_ARG4;
+		if (argnum == 2)
+			return R_ARG2;
+		else {
+			// No R_ARG4, but IF/ELSE may get the axe (CC #2077)
+			DS_RET_VALUE(D_ARG(argnum));
+			return R_RET;
+		}
 	}
 
 }
--- a/src/core/n-control.c
+++ b/src/core/n-control.c
@@ -593,7 +593,7 @@ got_err:
 {
 	REBCNT argnum = IS_FALSE(D_ARG(1)) ? 3 : 2;
 
-	if (IS_BLOCK(D_ARG(argnum))) {
+	if (IS_BLOCK(D_ARG(argnum)) && !D_REF(4) /* not using /ONLY */) {
 		DO_BLK(D_ARG(argnum));
 		return R_TOS1;
 	} else {
@@ -638,7 +638,7 @@ got_err:
 	} else
 		if (IS_FALSE(cond)) argnum = 4;
 
-	if (IS_BLOCK(D_ARG(argnum))) {
+	if (IS_BLOCK(D_ARG(argnum)) && !D_REF(5) /* not using /ONLY */) {
 		DO_BLK(D_ARG(argnum));
 		return R_TOS1;
 	} {
@@ -790,7 +790,7 @@ got_err:
 ***********************************************************************/
 {
 	if (IS_FALSE(D_ARG(1))) {
-		if (IS_BLOCK(D_ARG(2))) {
+		if (IS_BLOCK(D_ARG(2)) && !D_REF(3) /* not using /ONLY */) {
 			DO_BLK(D_ARG(2));
 			return R_TOS1;
 		} else {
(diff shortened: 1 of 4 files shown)


src/core/n-control.c
```diff
@@ -627,15 +631,20 @@ enum {
...
 		if (IS_FALSE(cond)) return R_NONE;
 	} else
-		if (IS_FALSE(cond)) body = D_ARG(4);
+		if (IS_FALSE(cond)) argnum = 4;
+
+	if (IS_BLOCK(D_ARG(argnum))) {
+		DO_BLK(D_ARG(argnum));
+		return R_TOS1;
+	} {
+		return argnum == 2 ? R_ARG2 : R_ARG4;
```
earl: I don't think it's worth it to introduce `R_ARG4` for this single use. You could use `DS_RET_VALUE(D_ARG(argnum)); return R_RET;` (off the top of my head) instead.

src/core/n-control.c
```diff
@@ -627,15 +631,26 @@ enum {
...
+		DO_BLK(D_ARG(argnum));
+		return R_TOS1;
+	} {
+		if (argnum == 2)
+			return R_ARG2;
+		else {
+			// No R_ARG4, but IF/ELSE may get the axe (CC #2077)
+			DS_RET_VALUE(D_ARG(argnum));
+			return R_RET;
+		}
```
earl, line 651: You can simplify lines 645 thru 651 to:

```
DS_RET_VALUE(D_ARG(argnum));
return R_RET;
```

hostilefork, line 651: Thought it was better to leave the R_ARG2 return form in the code, as we seem to all agree that IF/ELSE can die... so the lines will reduce to the even briefer `return R_ARG2;`

hostilefork:
I've added the /ONLY refinement to suppress block evaluation and just return the raw block.  Here are some examples of what the new behavior would do:

```
>> str: either (10 < 20) {Lesser} {Greater}
== "Lesser"

>> str: either 10 > 20 {Lesser} {Greater}
== "Greater"

>> val: either 10 < 20 1 + 2 3 + 4     
== 3

>> val: either 10 > 20 1 + 2 3 + 4 
== 7

>> result: either (10 < 20) (neg-offset: -1) (pos-offset: 1)
== -1
>> neg-offset
== -1
>> pos-offset
== 1

>> result: either 10 < 20 [ten less rule] [ten not less rule]
** Script error: ten has no value
** Where: either
** Near: either 10 < 20 [ten less rule] [ten not less rule]

>> result: either/only 10 < 20 [ten less rule] [ten not less rule]
== [ten less rule]
```

I realize people coming from other languages might find this to be a weird spectrum of possibilities for what you can do with conditional operations.  And there is a possibility that people would (for instance) try the summing examples without observing that it would run both computations regardless of the condition.

But that's just like everywhere else in Rebol.  Best to learn from the start.  Increasingly, I think Rebol+Red have a job to push back and show how uniform and flexible the evaluator is...drilling in the logic repeatedly and how much freedom people have.

And one of those freedoms is from the "No Choice But Ugly" that seems pervades every other language.  The current lack of an /ONLY option means that if you want to return blocks by value you wind up with things like:

```
>> result: either 10 < 20 [[ten less rule]] [[ten not less rule]]
== [ten less rule]
```

Which I think is uglier, with those doubling up of brackets.  It ceases to look like a smooth written language.  Similarly, I don't really like:

```
>> str: either 10 < 20 [{Lesser}] [{Greater}]
== "Lesser"
```

Instead, I prefer the look of:

```
>> str: either (10 < 20) {Lesser} {Greater}
== "Lesser"
```

I think it's about choice.  And Rebol has ripped away the safety net of boilerplate almost everywhere in the system to enable that choice.

It would be interesting to try working with this option on master and seeing what the tradeoffs might be in new bugs or frustration vs. clarity.  Not sure how often it would cause a weird confusing condition.  But ever since thinking of this, I am constantly finding cases where I would apply it and feel happier about the way the code looked.

My opinion is that this very example is a great demonstration for people learning the nuances of Rebol from the outset, and the choices of expression.  (Also in the "don't cater to mediocrity" department: IF/ELSE should die, and there seems to be consensus on that...)

PR #171, opened 2013-10-05
Fix Collect_Object to not overwrite memory BUF_WORDS does not own.

ladislav:
Without this fix, Collect_Object was causing crashes in 64-bit R3.
No regressions in Linux (0.4.4)


ladislav:
Merged to community.

PR #172, opened 2013-10-05
Make word lists of frames bare as suggested.

ladislav:
Based on #171
Tested in Linux (0.4.4) with no regression.


ladislav:
Merged to community.

PR #174, opened 2013-10-07
Add /ONLY refinement to PRINT, deprecate PRIN (CC #2070)

hostilefork:
This implements [CC #2070](http://curecode.org/rebol3/ticket.rsp?id=2070), and deprecates PRIN to a mezzanine-level function.  Ultimately it should be relocated to R3/Backward.

As the ticket points out, a PRINT variant that omits spaces between block items--in addition to omitting the newline--helps mitigate cases where PRIN might be called several times in succession.


hostilefork:
I've realized that fixing this kind of issue is not on the radar for the project.

PR #176, opened 2013-10-14
[JOIN, REJOIN] mezz => [ADJOIN, JOIN] natives (CC #2079)

hostilefork:
This implements [CC #2079](http://curecode.org/rebol3/ticket.rsp?id=2079) by replacing the mezzanines JOIN and REJOIN with natives ADJOIN and JOIN respectively.

Similar to FUNCT => FUNCTION, this is a change where a more popular construction retakes the cleaner name of a less popular construction, that was previously using that name.  (See `%tools/workarounds.r` for how sensing the number of arguments with `words-of` is used to ensure older r3-make binaries can continue to build the bootstrap.  )

For the moment, REJOIN is kept equivalent with JOIN.  The relative unpopularity of the previous JOIN (compared with REJOIN, at least) probably makes it unnecessary to be overly cautious about phasing the change.  Although if one were to be cautious, they could test a codebase to find JOIN instances by unsetting it (same w/FUNCTION).

Performance effects vary, but new JOIN has less overhead.  It's obviously the case that if significant work is happening in the evaluation of the expressions being reduced, the performance gain will become marginal by comparison.

Before:

```
>> delta-time [loop 10000 [rejoin [http://example.com %/number/ {=} "30"]]]
== 0:00:00.022506

>> a: [10 + 20] delta-time [loop 10000 [rejoin [http://example.com %/number/ {=} a]]]
== 0:00:00.025538
```

After:

```
>> delta-time [loop 10000 [join [http://example.com %/number/ {=} "30"]]]
== 0:00:00.009667

>> a: [10 + 20] delta-time [loop 10000 [join [http://example.com %/number/ {=} a]]]  
== 0:00:00.017016
```

Both ADJOIN and JOIN run through the same C function.  But in order to avoid the overhead of doing two REDUCE operations in JOIN, a parameter to the common function had to be added to indicate the `rest` value has already been reduced.  _(Whether this should be exposed as an actual /ONLY refinement is something that can be a consideration.)_

The best feature of this is not the performance, but the edging out of the RExxx conventions for _"REducing variant of XXX"_.  That convention leads to a host of extremely poor names like "REFORM" and "REJOIN"... and if taken to logical conclusions would produce more monsters like "REINSERT" and "RECHANGE".

### DIFF
--- a/src/boot/natives.r
+++ b/src/boot/natives.r
@@ -92,8 +92,8 @@ compose: native [
 	value "Block to compose"
 	/deep "Compose nested blocks"
 	/only {Insert a block as a single value (not the contents of the block)}
-	/into {Output results into a block with no intermediate storage}
-	out [any-block!]
+	/into {Output results into a series with no intermediate storage}
+	out [any-block! any-string! binary!]
 ]
 
 context: native [
@@ -251,8 +251,8 @@ reduce: native [
 	/no-set {Keep set-words as-is. Do not set them.}
 	/only {Only evaluate words and paths, not functions}
 	words [block! none!] {Optional words that are not evaluated (keywords)}
-	/into {Output results into a block with no intermediate storage}
-	out [any-block!]
+	/into {Output results into a series with no intermediate storage}
+	out [any-block! any-string! binary!]
 ]
 
 repeat: native [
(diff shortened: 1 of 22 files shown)


hostilefork:
Closed, although perhaps someone might want to revisit the native-ization of JOIN and REJOIN someday, as it would speed things up.

Also, I'm not quite sure why closing this looks to have closed stuff I did on another branch which let strings and binaries be targets of REDUCE/INTO and COMPOSE/INTO, but it did.

PR #177, opened 2013-11-13
Add a /only refinement to wait

zsx:
Such that it only checks events from the ports passed in as the
argument. This could be used to ease sychroneous communication and used
for implementing a pause when the port list is empty (in case of no GUI events).


carls:
Call for community comments on this enhancement of WAIT. (I need to start doing this more to get inputs and discussion. Too bad there's not some way to mark these in github... or, maybe there is?)

hishamhm:
@carls you can use the Labels feature for that

PR #182, opened 2014-02-17
Fix a regression in PARSE caused by 8db79a9

earl:
This fixes an operator precedence typo in the check for "not COPY or SET" (introduced by 8db79a9 from PR #165) which causes a few regressions in PARSE.

Tested under Linux-x86: eliminates the 25 regressions.


hostilefork:
My bad.  As noted in chat, it's been quite a while (5 months ago)... so I don't know whether I either neglected to test that one, or did but misread the test results somehow.  :-/  My use of the test scripts is somewhat ad-hoc so it's actually equally likely that I just made a mistake in my executable choices or choice of logs to diff as it is that I forgot to do it.

But thanks for the catch/patch!

PR #183, opened 2014-02-17
FIND-ALL speed optimization

BrianHawley:
Requested by Ladislav in http://issue.cc/r3/2005

### DIFF
--- a/src/mezz/mezz-series.r
+++ b/src/mezz/mezz-series.r
@@ -501,15 +501,18 @@ split: func [
 	]
 ]
 
-find-all: function [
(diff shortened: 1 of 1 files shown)


BrianHawley:
Because of http://issue.cc/r3/539 this function needs to be made into a native, but it doesn't hurt to get this speedup in the meanwhile.

carls:
Is the indentation non-standard on this code? What's a good way to fix this?

earl:
The indendation of `find-all` in `src/mezz/mezz-series.r` was _previously_ non-standard. This pull fixes this as well.

PR #184, opened 2014-02-20
Fix evaluation of lit-words, lit-paths and set-paths

BrianHawley:
- Make inline evaluation of lit-word assigned to word return lit-word
- Make explicit evaluation of lit-path return path
- Make explicit evaluation of set-path trigger same error as set-word
- Fixes indentation style inconsistency in n-control.c
- Changed conversion in sys-ports.r from implicit to explicit

Fixes http://issue.cc/r3/1434 and http://issue.cc/r3/1883


BrianHawley:
Ticket 1434 requested again as http://issue.cc/r3/2101 so mark that too.

earl:
Builds fine, but a build with those changes merged seems to be rather broken:

```
$ make/r3
** access error: missing port scheme: 'system

>>
```

BrianHawley:
Fixed. There was an error in sys-ports.r that was hidden by the old behavior.

PR #185, opened 2014-02-20
Get rid of if/else

BrianHawley:
Per request: http://issue.cc/r3/2077


BrianHawley:
Bonus, it makes `if` faster and take less stack space!

earl:
Tested on Linux x86: good to go.

earl:
r+

PR #187, opened 2014-02-20
Ordinals going in the other direction

BrianHawley:
See http://issue.cc/r3/1983 for details.

Also added replacements of these to tools in case they're needed, as they were in some existing (commented-out) code.


BrianHawley:
This is part one of the great index compromise we worked out last year. Two more parts coming.

BrianHawley:
If #140 goes through first I'll have to adjust this. If this goes through first #140 would need adjusting.

ladislav:
I am not sure the new functions will be popular enough to justify their introduction.

BrianHawley:
The purpose of these functions is to make differences in the model of less-than-one `pick` indexes between Rebol derivatives and versions not important in portable code. The other parts of the compromise were to add 0-based `pickz` and `pokez` native wrappers for `pick` and `poke`, and to change the indexing model of `pick`, `poke` and path expressions back to R2-style 1-based, no-0 indexing.

The goal of the first two is to change _all_ less-than-one uses of `pick` and `poke` to use the backwards ordinals, `pickz` or `pokez` instead. Once that code conversion has happened, we will be free to unify the behavior of `pick` and `poke` indexing with Red, so code can be portable across all the Rebol-like languages. They would go into R2/Forward as well, for those still using Rebol 2.

This is the only combination of features and changes that I could get the Rebol and Red people to agree on (that included you, Ladislav, but that was a while ago). If you want to have the debate again, that's fine, but make sure the Red people agree with you before you continue.

I'm working on `pickz` and `pokez` now (see [CC#613](http://issue.cc/r3/613)). Should I add the suggested `indexz?` as well?

ladislav:
Well, no, I do not want to debate the functions. I am just unsure whether they will be "popular enough" for general inclusion. (but that is nothing I can decide or influence)

BrianHawley:
Sorry, I missed out on mentioning the compromise plan. Comment updated.

BrianHawley:
"but that is nothing I can decide or influence"

Funny, you were the primary influence for the 0-based model, the reason for `pickz` and `pokez` :)

You were in favor of keeping the contiguous index space for path access, even if `pick` and `poke` were changed. That wasn't going to be agreed on by the 1-based non-contiguous advocates, like the Red developers, so it wasn't going to work. As a compromise for you, I agreed to push to get `pickz` and `pokez` added so people like you who needed to do index math would have available to them the best possible solution for that (0-based contiguous). Yay, everyone wins!

ladislav:
I am sorry if I wrote my above comment so unclearly that it did not communicate my inability to decide or influence how _popular_ the functions introduced in this PR are going to be. Nevertheless, I still think that it makes a little sense to introduce unpopular and unused functions to the interpreter.

BrianHawley:
Agreed. But the point of these is for us to _encourage their use_, instead of the use of `pick` with less-than-one indexes, because the latter is _not portable_, and probably should be considered subject to change until it _is_ portable.

However, back to the point of "it makes a little sense to introduce unpopular and unused functions to the interpreter", it is still the plan to make it possible to make a Rebol without all of that stuff, and possible to make a version with all of that included, but optionally loaded. Which is why I've been working on cleaning up and finalizing the module system, since the plan is to put this stuff into modules.

BrianHawley:
Curecode issue documenting the compromise plan: http://issue.cc/r3/2117

PR #188, opened 2014-02-21
-fixed the infamous #9910 crash

cyphre:
the #9910 crash was caused by ports in pending requests being GCed while the system port or app code could still access the pending events with port refs

To test the crash you can run following gist: https://gist.github.com/cyphre/9122095

related addition:
-added Host_Lib->devices reference to be able access host Devices array from core dll

And thanks must go to Ladislav as well as he assisted me when looking up the best solution for the fix.

### DIFF
--- a/src/core/m-gc.c
+++ b/src/core/m-gc.c
@@ -81,6 +81,10 @@
 #include "sys-core.h"
 #include "reb-evtypes.h"
 
+#ifdef REB_API
+extern REBOL_HOST_LIB *Host_Lib;
+#endif
+
 //-- For Serious Debugging:
 #ifdef WATCH_GC_VALUE
 REBSER *Watcher = 0;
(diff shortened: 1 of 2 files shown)


earl:

earl:
r+

carls:
I've accepted this change because it fixes the bug, but doesn't it short circuit the architecture of the isolated core? We should find another way, probably by adding a host device function of some kind, but I'd need to study it closer. Maybe someone has some ideas on it already?

earl:
Two ways I thought about:
1. We could add an `OS_Walk_Device_Ports` function to the host interface. This function would traverse the pending lists of all devices. It would not mark the ports of pending requests on it's own, as that would leak the marking logic from core to host. But rather, it would invoke a function pointer with one argument: the REBSER\* for the port. GC would call OS_Walk_Device_Ports with a simple "mark this series" callback.
   
   (So the full signature would be: `void OS_Walk_Device_Ports(void(*handle_port)(REBSER*))`.)
2. Maybe manual management by judicious use of the already-existing `RL_Protect` in the core interface could fix this issue as well. Mark the corresponding port as GC-protected during device request attachment and unprotect on detaching a request (i.e. in `Attach_Request` and `Detach_Request`).

PR #189, opened 2014-02-21
Non-copying FUNC-BOOT for all mezzanines + explicitly distinct from FUNC

hostilefork:
The Mezzanine code (mostly) used a non-copying variant of FUNC, [simply defined](https://github.com/rebol/rebol/blob/4c32cc73b4641f8f6831b3ac549b5f5084c9e159/src/mezz/base-funcs.r#L19) as this:

```
func: make function! [[
    ; !!! This is a special minimal FUNC for more efficient boot. Gets replaced later in boot.
    {Non-copying function constructor (optimized for boot).}
    spec [block!] {Help string (opt) followed by arg words (and opt type and string)}
    body [block!] {The body block of the function}
][
    make function! reduce [spec body]
]]
```

Not copying creates a lot of potential for problems.  But a simple example shows why we don't define FUNC this way for the average user:

```
body: [
    value: {}
    append value {text}
    return value
]

x: func-boot [] body
insert body [print "Entering y function"]
y: func-boot [] body

x
y
```

Since the func-boot generator does not copy the body, this would mean the body winds up shared by both functions.  So calling x prints "Entering y function" also, and they share the same `value` string:

```
Entering y function
text
Entering y function
texttext
```

However, not copying saves on time and memory.  As long as you're not passing the same body to multiple functions, it should be able to work.

When the mezzanine was done with its loading, it would switch over to an implementation of FUNC that deep copied... [saving the old version into FUNCO](https://github.com/rebol/rebol/blob/08eb7e84cdadf450727501bb8ec02fad45696959/src/mezz/mezz-tail.r#L14)

There are a few problems with the way things were done:
- It was not easy to know by looking at code which variant of FUNC was in effect.  This made the usage of some mezzanines from within the mezzanine dangerous (such as COLLECT).  They would only be safe to use at the point where the changeover to a copying FUNC happened.
- Not all mezzanines followed the convention, and there were FUNCTION instances being used in various places... even ones with no set-words in them.  Given that it is hand-maintained system code, the time to scan the bodies for locals is probably not well spent.  _(There are also "word not bound to a context" errors to help catch any missed LOCALs.)_
- Most users would not be able to use FUNCO safely so exporting it is probably a bad idea.

So instead of changing the definition of FUNC in midstream, it seems much better to be explicit about which one the code is using.  FUNC-BOOT seemed like a more informative name than FUNCO.

FUNC is defined up front to give it an "early slot" (still not sure why that matters).  However it gives an error message if you use it too early, and returns #[unset!]

I changed it to not define FUNCTION until `%mezz-func.r`, which is currently the last thing that runs before `%mezz-tail.r`.  Any mezzanine functions that need to invoke FUNCTION at boot time would have to thus be after `%mezz-func.r`.  I turned all the existing FUNCTION instances prior to that into FUNC-BOOT with appropriate LOCALs.  Extra eyeballs to see what I might have missed are helpful; though I've done enough to keep the test suite from complaining.

The copying FUNC is also given its real definition in `%mezz-func.r`.

In `%mezz-tail.r`, the FUNC-BOOT is defined to produce an error message when invoked.  This prevents its use by users...while still giving a meaningful message for any code taken out of the mezzanine that is not adapted to use FUNC instead of FUNC-BOOT.

To help catch the kind of problematic cases that would mean you can't use FUNC-BOOT (e.g. modifying bodies) I did a PROTECT/DEEP on the spec and body passed in.  Outside of needing a couple of COPY operations on help strings that were trimmed in place (which probably should have only been done once anyway), it didn't need much changing.

Now that the FUNC-BOOT calls can be seen explicitly at the sites where they are used, it should be possible to review them for any cases where non-copying semantics would be bad.  And if you try to use something like COLLECT before the real FUNC is defined you'll get a runtime error.  This brings all that out into the open instead of having to keep track of it in one's head _(...if someone editing the code even realized it was going on in the first place!!)_

### DIFF
--- a/src/mezz/base-debug.r
+++ b/src/mezz/base-debug.r
@@ -16,7 +16,7 @@ REBOL [
 	}
 ]
 
-probe: func [
+probe: func-boot [
 	{Debug print a molded value and returns that same value.}
 	value [any-type!]
 ][
@@ -24,7 +24,7 @@ probe: func [
 	:value
 ]
 
-??: func [
+??: func-boot [
 	{Debug print a word, path, or block of such, followed by its molded value.}
 	'name "Word, path, and block to obtain values."
 	/local out
@@ -55,14 +55,14 @@ probe: func [
 	:name
 ]
 
-boot-print: func [
+boot-print: func-boot [
 	"Prints during boot when not quiet."
 	data
 ][
 	unless system/options/quiet [print :data]
 ]
 
-loud-print: func [
+loud-print: func-boot [
 	"Prints during boot when verbose."
 	data
 ][
--- a/src/mezz/base-defs.r
+++ b/src/mezz/base-defs.r
@@ -36,7 +36,7 @@ use [word title] [
 			values      ["object or module"]
 			types title ["function"] ; title should include module Title too...
 		] ["function, object, or module"]] ; body, words
-		set word func
+		set word func-boot
 			reduce [title 'value]
 			compose [reflect :value (to lit-word! name)]
 	]
@@ -90,57 +90,57 @@ system/options/result-types: make typeset! [
 
 ;-- Create "To-Datatype" conversion functions early in bootstrap:
 
-any-block?: func [
+any-block?: func-boot [
 	"Return TRUE if value is any type of block."
 	value [any-type!]
 ][find any-block! type? :value]
 
-any-string?: func [
+any-string?: func-boot [
 	"Return TRUE if value is any type of string."
 	value [any-type!]
 ][find any-string! type? :value]
 
-any-function?: func [
+any-function?: func-boot [
 	"Return TRUE if value is any type of function."
 	value [any-type!]
 ][find any-function! type? :value]
 
-any-word?: func [
+any-word?: func-boot [
 	"Return TRUE if value is any type of word."
 	value [any-type!]
 ][find any-word! type? :value]
 
-any-path?: func [
+any-path?: func-boot [
 	"Return TRUE if value is any type of path."
 	value [any-type!]
 ][find any-path! type? :value]
 
-any-object?: func [
+any-object?: func-boot [
 	"Return TRUE if value is any type of object."
 	value [any-type!]
 ][find any-object! type? :value]
 
-number?: func [
+number?: func-boot [
 	"Return TRUE if value is a number (integer or decimal)."
 	value [any-type!]
 ][find number! type? :value]
 
-series?: func [
+series?: func-boot [
 	"Return TRUE if value is any type of series."
 	value [any-type!]
 ][find series! type? :value]
 
-scalar?: func [
+scalar?: func-boot [
 	"Return TRUE if value is any type of scalar."
 	value [any-type!]
 ][find scalar! type? :value]
 
-true?: func [
+true?: func-boot [
 	"Returns true if an expression can be used as true."
 	val ; Note: No [any-type!] - we want unset! to fail.
 ] [not not :val]
 
-quote: func [
+quote: func-boot [
 	"Returns the value passed to it without evaluation."
 	:value [any-type!]
 ] [
--- a/src/mezz/base-files.r
+++ b/src/mezz/base-files.r
@@ -16,21 +16,21 @@ REBOL [
 	}
 ]
 
-info?: func [
+info?: func-boot [
 	{Returns an info object about a file or url.}
 	target [file! url!]
 ][
 	query target
 ]
 
-exists?: func [
+exists?: func-boot [
 	{Returns the type of a file or URL if it exists, otherwise none.}
 	target [file! url!]
 ][ ; Returns 'file or 'dir, or none
 	select attempt [query target] 'type
 ]
 
-size?: func [
+size?: func-boot [
 	{Returns the size of a file.}
 	target [file! url!]
 ][
@@ -40,7 +40,7 @@ size?: func [
 	]
 ]
 
-modified?: func [
+modified?: func-boot [
 	{Returns the last modified date of a file.}
 	target [file! url!]
 ][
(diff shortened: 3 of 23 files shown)


BrianHawley:
"It was not easy to know by looking at code which variant of FUNC was in effect."

_The non-copying one was in effect_. The files are run in the order they are specified in `boot-files.r`, and the last file of the mezzanines is `mezz-tail.r`, where the non-copying `func` is replaced with the copying one. So, every file specified before that one uses the non-copying one. The only file that uses the copying `func` is the `prot-http.r` module (along with the rest of the protocol section).

The reason we don't use `collect` during the mezzanine creation process is that we don't use almost any mezzanines at all during the mezzanine creation process. We don't use recursive functions either, coincidentally, and try to use mostly natives and very simple mezzanines. The startup code is pretty tricky, you have to make things with the tools available at the time, which is why the files are loaded in the order specified. Also, it has to be fast and relatively low memory, which is why we don't use a copying `func` or functions like `collect`.

We don't need to have another name for the boot-time `func` because that word would have to be bound somewhere, and once it's bound there you can't get rid of that word without getting rid of the context it's bound to - and all of those words are bound to `sys` or `lib`. And we really don't need to keep the function around after the mezzanines are done loading, so it's just a word leak. The only way to avoid leaving that word in the context is to not define it in the first place.

On that note, we don't need to define `funco` either. Let's add it to the cleanup list.

As for the `protect/deep` idea, it's an interesting idea. We can do it without changing the name of the boot-time `func` too, just add the protection code to the that function as well.

"I changed it to not define FUNCTION until `%mezz-func.r`,"

We need to have `function`  defined early because many of the mezzanine functions are defined with `function` for development convenience. All uses of `function` without options should be converted to regular `func` form during the build process so the startup won't have the overhead of `function` during the startup time. Best of both worlds. Only the ones that use `function`'s options can't be translated to only `func` because `func` doesn't do what those options do, though in theory we could translate those to function building expressions if we wanted to (I'll look into that later).

Don't make it harder to write the mezzanines by making `function` not available, or by making us keep track of the use of the non-copying `func` when we don't need to do that.

OK, to-do list from this: get rid of `funco`; apply that `sys/load-header` change and a few others; see about having the builder code translate `function/with` use to function-building expressions; and check whether `closure` is being used and needs the same translation as `function`. Huzzah!

hostilefork:
> And we really don't need to keep the function around after the mezzanines are done loading, so it's just a word leak. The only way to avoid leaving that word in the context is to not define it in the first place.

It was nice for me to have it defined while debugging, so I could see the bodies of the functions that were trying to use it that shouldn't in the output.  But it could be unset and that would be fine also.

> We need to have function defined early because many of the mezzanine functions are defined with function for development convenience, then converted to regular func form during the build process so the startup won't have the overhead of function during the startup time, except for the ones that use function's options because they don't translate. Best of both worlds.

Put works-in-progress in %mezz-work.r between %mezz-func.r and %mezz-tail.r.  When you move them out of mezz-work and into an earlier file, then you convert them.  _That's_ best of both worlds; keeps a nice level of consistency on the "we're doing this for boot speed" thing.

> Don't make it harder to write the mezzanines by making function not available, or by making us keep track of the use of the non-copying func when we don't need to do that.

I'd phrase it as "don't make debugging any harder than it needs to be".  You identified a bug by off the cuff pointing out the lack of safety w.r.t. COLLECT.  Then you say "well, I can just keep track of it in my head".  But if the code is to be modified by more than one person then they need all the help they can get... both in the source and at runtime when they've done something they shouldn't.  Imagine how sneaky figuring out what went wrong could be.  This is cheap preventive medicine.

As for where to define FUNCTION and other helpers, I guess I don't care.  I don't know if I'd think that the mezzanine routines needed a special FUNC to load faster and save memory in the first place, though it sounds plausible...be a little more careful with the function, save a little bit.  But if you say _"we want the mezzanine to boot fast by writing code that uses a special version of func"_ and then say _"yeah, but, we don't really use it all the time... just, y'know if we feel like it"_ then I'll shrug and say it's very hard to codify best practices in development for something a team of people are working on with that kind of dithering.

My real problem here is with changing the semantics of a foundational function and then switching what the name binds to at runtime.  You've listed a class of bugs that could be caught and prevented.  It's worth worrying about.

BrianHawley:
"Put works-in-progress in %mezz-work.r between %mezz-func.r and %mezz-tail.r."

They are all works in progress, even after they're final. We don't want to make further maintenance more difficult either. We want these to continue to be written with `function` unless they can't be. The builder code should be able to translate it for us perfectly, there is no reason for us to do it manually.

"You identified a bug by off the cuff pointing out the lack of safety w.r.t. COLLECT."

_It's not a bug_. We don't use `collect` until it's safe, but that's not why. We don't _want_ to use `collect` in the startup code _at all_ because of its overhead. It's a user function, not a system function. We can't use the `to-*` type conversion functions either because they aren't bound until later, and that's not a bug either. Mezzanines are a bit advanced to write, but it could be a lot worse if we didn't control the build process.

The mezzanines are built efficiently and have to follow the code structure that allows them to be built. The way we manage to use the nice user-space tools at all is that a lot of the building process is done at build time rather than at startup time. That is how we are able to use the nice toys at all, and using the nice toys makes it easier to write this stuff so let's not stop using them.

"But if the code is to be modified by more than one person then they need all the help they can get..."

That's the job of code comments and other documentation (which are also stripped out by the build process). We add comments so we can do the stuff we need to to make the process work well, without having the overhead of that at runtime.

Rebol has to be hand-optimized due to its nature, but it _can be_ hand optimized for the same reasons. Sometimes it takes some tricky code to implement this simplicity, and that's fine.

If we used `func-boot` in the code, it would be easy enough to post-process that name away too, changing things to the boot-time `func` during the build process, so that's not really a problem I guess if you want to put that in there. I don't yet buy that it's necessary or a good idea though.

BrianHawley:
Note: The bootstrap code currently doesn't translate `function` calls to `func`, so I made #196 to change all mezzanine declarations from `function` to `func`, except the `secure` function. It also includes all of the bug fixes from this PR, and doesn't have missing locals or bad syntax. This PR should be closed, or rebased on #196 after it's merged.

PR #190, opened 2014-02-22
Android

giuliolunati:
Minor changes to compile for android


BrianHawley:
Command line Android is basically Linux, but built against a different libc, Bionic. Since it's a different libc we need to have new platform numbers for that build, so we can make prebuilt binaries and have them work. The 0.4.3 platform isn't the right one to be using here, that is for 32-bit x86 Linux builds against glibc version 2.5, so code that requires that specific platform will fail.

We need two new minor platform variant numbers, one for 32-bit x86 Linux built against Bionic, and one for 32-bit ARM Linux built against Bionic. And then maybe two more for 64-bit.

BrianHawley:
The build-on-Android method documented requires AIDE, and rooting your phone. We should also include Terminal IDE instructions, so you don't have to root your phone unless you want to install Rebol for the whole system. Just building Rebol shouldn't require root access.

BrianHawley:
#191 supersedes this one, so this one can be closed. Thanks @giuliolunati and @earl!

carls:
Superseded by #191

PR #193, opened 2014-02-27
Fix word hash collision and function parameter type mismatches

BrianHawley:
- Fixes the word comparison problem from http://issue.cc/r3/2113
- Clean up REBINT vs REBCNT type mismatches in UTF decoding.


earl:

earl:
r+

BrianHawley:
Note: This bug was one of the blockers preventing Red's bootstrap being ported from Rebol 2 to 3.

PR #194, opened 2014-02-27
DO-CODEC related fixes

cyphre:
This pull request fixes the regression crash introduced by https://github.com/rebol/rebol/pull/156 which caused TEXT type decoding operation non-functional.
It also fixes the crash when encoding image using PNG codec ([CC#2040](http://issue.cc/r3/2040)).

Here is a link to Gist that includes all basic DO-CODEC tests for possible reveal of regression:
https://gist.github.com/cyphre/9254624

### DIFF
--- a/src/core/u-zlib.c
+++ b/src/core/u-zlib.c
@@ -287,6 +287,8 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     }
     if (strm->zfree == Z_NULL) strm->zfree = zcfree;
 
+	if (!strm->checksum) strm->checksum = adler32;
+	
     if (level == Z_DEFAULT_COMPRESSION) level = 6;
 /* #ifdef FASTEST */
 /*     level = 1; */
--- a/src/core/n-system.c
+++ b/src/core/n-system.c
@@ -443,19 +443,21 @@ err:
 	case CODI_CHECK:
 		return R_TRUE;
 
-	case CODI_BINARY:
-	case CODI_TEXT:
+	case CODI_BINARY: //used on encode
+	case CODI_TEXT: //used on decode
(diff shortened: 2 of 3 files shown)


BrianHawley:
Looks good.

Those tests have clearly shown that our test suite needs support for fixtures. No tests should refer to external resources, like a web site. We need to be able to include the files we're testing with the tests.

Good tests though :)

earl:
Thanks for the quick fix, Cyphre. Also nice that the PNG crash could be so easily fixed as well.

Tested on Linux x86 (4.4): no regressions, 2 previous crashes fixed (the PNG and the TEXT crash).

earl:
"We need to be able to include the files we're testing with the tests."

That's already possible even with the current test system. We should incorporate the tests proposed in the Gist into the test suite. I'll contact you directly, @cyphre, to assist in doing so. (And I think @ladislav would be happy to assist as well.)

src/core/u-zlib.c
```diff
@@ -287,6 +287,8 @@ int ZEXPORT deflateInit2_(strm, level, method, windowBits, memLevel, strategy,
     }
     if (strm->zfree == Z_NULL) strm->zfree = zcfree;
 
+	if (!strm->checksum) strm->checksum = adler32;
```
earl, line 290: This line is a tab-indented in a file using _predominantly_ space-based indents (but it's using mixed indents in general, already).

I don't think that minor whitespace quirk should block this otherwise important crash fix, though. But we should probably do a clean up pass thru this file separately.

earl:
r+

src/core/u-zlib.c
BrianHawley, line 290: Yeah, get it in ASAP, clean up whitespace later.

PR #195, opened 2014-03-02
fix ticket1457 parse to bitset

giuliolunati:


earl:
This implements bitset handling for TO/THRU in PARSE ([CC#1457](http://issue.cc/r3/1457)).

PR #196, opened 2014-03-04
Use FUNC instead of FUNCTION to speedup startup

BrianHawley:
Translated uses of FUNCTION to FUNC, to lower overhead during Rebol
startup code. Guess FUNCTION wasn't pretranslated by the bootstrap.

Also, fixed a few literal series modifications that could be copies.

### DIFF
--- a/src/mezz/base-files.r
+++ b/src/mezz/base-files.r
@@ -172,9 +172,10 @@ split-path: func [
 	reduce [dir pos]
 ]
 
-intern: function [
+intern: func [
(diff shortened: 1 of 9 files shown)


BrianHawley:
If we want these to remain as `function` for now, for ease of development, then I can modify this PR to only include the bug fixes. Perhaps we'll decide to have the bootstrap code preprocess `function` calls to `func`, so this kind of thing won't be necessary.

PR #197, opened 2014-03-04
Get rid of many compiler warnings

BrianHawley:


BrianHawley:
Sorry, that PR close was due to a UI bug in Github.

PR #199, opened 2014-03-04
Make OBJECT no longer initialize to none

BrianHawley:
See http://issue.cc/r3/2118 for details.


BrianHawley:
Closing, needs to be more thorough.

Oldes:
Mentioned link above is here: https://www.curecode.org/rebol3/ticket.rsp?id=2118 which was imported to: https://github.com/rebol/rebol-issues/issues/2118

PR #200, opened 2014-03-04
Add SCOPE function

BrianHawley:
See http://issue.cc/r3/2119 for details.

### DIFF
--- a/src/mezz/mezz-control.r
+++ b/src/mezz/mezz-control.r
@@ -26,3 +26,10 @@ launch: func [
 	if arg [append args arg]
 	either wait [call/wait args] [call args]
 ]
+
+scope: func [
+	"Evaluates a block with all set-words as locals."
+	body [block!] "Block to evaluate"
+][
+	do bind/copy/set body make object! 0
+]


BrianHawley:
There is some debate in CureCode over the name. Though `scope` is a good name for the intended use of this function - delineating a smaller section of code with words local to that code section, similar to "scopes" in languages with lexical scoping - it has been pointed out that some Rebol people are relatively proud of not having scopes in the language. Also, to people who don't have a CS background, the term "scope" might be confusing.

So, we need a name for this function that is easy to understand, and short enough to use often. Keep in mind that the goal is to completely replace _all_ uses of the `context` function where the object value returned from `context` is not used - when it's being used for making local variables. This may also be used in many cases where `use` is used now, because of the convenience of locals-gathering, and possibly improved performance (haven't tested this yet, but it seems likely).

So, we need a nice sounding, understandable, short name with positive or neutral connotations. Any ideas? Otherwise, `scope` is neutral to most people.

BrianHawley:
For Rebol people who know about "definitional scoping", and that we really don't have scopes, per-se, `scope` would be an in-joke given that it's purpose is to create one of those definitional scopes that only pretend to be lexical scopes. Does the joke merit the name?

BrianHawley:
Another suggestion: Name this function `context` and rename the existing `context` function(s) to `object` (in a new version of #199). This would then be an enhancement of the `context` function.

Then, the original `context` function could be put in the backwards-compatibility module for those who need it, though we would prefer that `object` be used for clarity. Another R3-Backwards candidate.

BrianHawley:
Name settled, but this PR won't work anymore. Closing. I'll make a new PR.

Oldes:
The link with details mentioned above is: https://www.curecode.org/rebol3/ticket.rsp?id=2119

PR #203, opened 2014-03-05
Restructure platforms list for easier extensibility & cleanup whitespace and libstdc++ linking

earl:
Primarily, this is a minor restructuring of `src/boot/platforms.r` to make it easier to maintain (and merge).

On top of that, the platform build specification is consistently reformatted. Previously, while always using hard tabs for indentation, a complete mixture of tabs and spaces was used for alignment. This is cleaned up to always use hardtabs both for indentation and alignment.

Finally, the unneccesary linking against libstdc++ for FreeBSD and OpenBSD is dropped. Test builds on both systems succeeded with that linkage, and the resulting binaries work just fine.

### DIFF
--- a/src/boot/platforms.r
+++ b/src/boot/platforms.r
@@ -14,18 +14,82 @@ REBOL [
 	}
 ]
 
-Amiga		[1 m68k20+ 2 m68k 3 ppc]
-Macintosh	[1 mac-ppc 2 mac-m68k 3 mac-misc 4 osx-ppc 5 osx-x86]
-Windows		[1 win32-x86 2 dec-alpha]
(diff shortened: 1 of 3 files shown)


earl:
Footnote 1: GitHub displays hard tabs as 8-spaces wide by default. So the indendation/alignment changes _look_ like they mainly make things worse, even though that's not actually the case.

Footnote 2: `src/tools/systems.r` is actually a pretty good example why I personally still don't consider hardtabs to be a good idea. But that's a discussion to be had elsewhere (such as PR #56). The previous formatting was _inconsistent_ in any case. With this change, it is at least now consistently using one method.

BrianHawley:
Consistency is nice. I prefer hard-tabs for indentation, spaces for inline alignment, just because of inconsistencies like the one displayed in `systems.r`. But, whatever we do let's do it consistently.

PR #205, opened 2014-03-06
TO-VALUE function

BrianHawley:
This fills a hole in the basic evaluation model: converting non-values to values. This is how you turn an unset value that doesn't have to be considered erroneous into a nice safe none; all else passed through. Filters out the unset values from the data flow.

See http://issue.cc/r3/2003 for details.


earl:
r+

PR #208, opened 2014-03-08
Fix PARSE position capture combined with SET or COPY

earl:
The recent extension to allow set-word! usage in SET and COPY rules introduced a regression: set-word!s (for position capture) in the rule immediately following a `SET word` or `COPY word` rule cause a match failure. For example:

```
>> parse [x] [set val pos: word!]
== false  ;; Expected: true
```

This commit fixes this issue ([CureCode issue 2130](http://issue.cc/r3/2130)), a regression caused by 8db79a9 (PR #165).

Corresponding tests have been added to the test suite (rebolsource/rebol-test@fe8df53).

### DIFF
--- a/src/core/u-parse.c
+++ b/src/core/u-parse.c
@@ -794,11 +794,8 @@ bad_target:
 
 			} else { // It's not a PARSE command, get or set it:
 			
-				// word: - if not the target of a COPY or SET operation, this will
-				// default to setting a variable to the series at current index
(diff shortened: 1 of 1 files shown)


BrianHawley:
Looks good to merge if the new tests pass.

earl:
Tested on Linux x86: no regressions, 4 progressions (the newly added bug-specific tests).

rebolek:
Wonderful. Thanks! :)

PR #210, opened 2014-03-14
Add support for APPLY :DO [:function]

BrianHawley:
APPLY :DO [:function] handles the first argument block (the one to DO)
but doesn't handle the second (the one to the function). This change
will allow you to handle the second argument block with a chained
call to APPLY, or to not do the reval at all if you like.

See http://issue.cc/r3/2135 for details.

### DIFF
--- a/src/core/n-control.c
+++ b/src/core/n-control.c
@@ -293,6 +293,7 @@ enum {
 ***********************************************************************/
 {
 	Apply_Block(D_ARG(1), D_ARG(2), !D_REF(3)); // stack volatile
+	VAL_CLR_OPT(DS_TOP, OPTS_REVAL);
 	return R_TOS;
 }
 


ladislav:
The proposed behaviour looks unintuitive.

BrianHawley:
(Answer copied from CureCode, where you made a similar comment there. Pick one venue, Ladislav.)

Well, if you want it all in one block then the equivalent call would be 

```
>> apply :do [:add none none none none 1 1] 
== 2 
```

Remember `do`'s optional arguments, which are normally padded in. And you wouldn't be able to block the reval altogether, which is the primary reason for this proposal. 

If you wanted something more like `do`, you could call `do` directly. You use `apply` to be able to have more control over the evaluation; `apply` is not `do`.

(End of copied answer.)

What do you mean by "unintuitive", specifically? And keep in mind, this is `apply`, so not providing the arguments to the function in a block is not a valid option. Perhaps you have an alternate solution that handles both these goals:
- Make it possible to block the reval evaluation. We don't have another method for this, and need one.
- Allow you to provide a block of arguments to the function, so you can control their evaluation too.

If this is a criticism of `apply`'s arguments being positional, that is a completely separate issue. I agree that those positional arguments are unintuitive, but there has been no suggestion yet for an alternative that serves the function-wrapper purposes of `apply`. Do you have a suggestion for that? If so, make it in CureCode, in a separate ticket.

PR #213, opened 2014-03-23
Fix ticket 2138

giuliolunati:
PARSE string using tag! doesn't consider the < and > when used with to block!


Oldes:
PR used in: https://github.com/Oldes/Rebol3/commit/b940cbec774f87d2a18c06b53baa607126df5186

PR #215, opened 2014-04-01
Fix 2140 long cmdline

giuliolunati:


Oldes:
Used in: https://github.com/Oldes/Rebol3/commit/f27e71964a4cf9b153b25d919b90fd35f291560a

Although I'm quite not sure if the size was not limited by design (to prevent potential DOS attacks)

PR #217, opened 2014-04-05
Add PAIR! support for REPEAT

earl:
This change allows PAIR! values as value argument of REPEAT. This
enables convenient traversal over a 2D space:

```
>> repeat p 3x3 [print p]
1x1
2x1
3x1
1x2
2x2
3x2
1x3
2x3
3x3
```

Similar to REPEAT over decimals, REPEAT over pairs implies a start of
1.0x1.0 and a default bump of 1.0 for each component. The first pair
component ("X") is varied first, which is in line with pair!-based
indexing into image!s.

For the sake of completeness, PAIR! support is also enabled in the
legacy C-style FOR:

```
>> for p 10x10 15x20 5x5 [print p]
10x10
15x10
10x15
15x15
10x20
15x20

>> for p 3x1 1x2 -1x1 [print p]
3x1
2x1
1x1
3x2
2x2
1x2
```

This implements [CureCode wish #1995](http://issue.cc/r3/1995).

### DIFF
--- a/src/boot/natives.r
+++ b/src/boot/natives.r
@@ -150,9 +150,9 @@ find-script: native [
 for: native [
 	{Evaluate a block over a range of values. (See also: REPEAT)}
 	'word [word!] "Variable to hold current value"
-	start [series! number!] "Starting value"
-	end   [series! number!] "Ending value"
-	bump  [number!] "Amount to skip each time"
+	start [series! number! pair!] "Starting value"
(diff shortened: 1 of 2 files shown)


earl:

Oldes:
Used in: https://github.com/Oldes/Rebol3/commit/2231a1db082cbb80649c53bb20c3f2476e4d949f

PR #220, opened 2014-05-09
Fix non-/wait CALL for POSIX platforms (cc#1861)

earl:
Currently, on POSIX platforms, `call/wait` and `call` _both_ block until the called program finishes. This PR fixes plain CALL to be non-blocking on POSIX platforms as well.

### DIFF
--- a/src/core/n-io.c
+++ b/src/core/n-io.c
@@ -523,18 +523,21 @@ chk_neg:
 **
 */	REBNATIVE(call)
 /*
+**		1 command     "An OS-local command line, quoted as necessary"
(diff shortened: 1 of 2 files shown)


earl:
Confirmed to build and work as expected on:
- Linux x86 (GCC 4.9.0)
- OS X x86 (clang 3.4)
- Haiku x86 (gcc 2.95.3).

zsx:
Looks good!

TGD-Consulting:
Confirmed to build and work as expected on: 

```
RPi (Linux libc6-arm)
```

PR #223, opened 2014-06-11
FEAT: Initial serial port implementation for POSIX and WIN32

kealist:
# Serial Implementation

This pull request includes two commits:
- The first commit is Carl's serial code implementation from the ReCode conference.
- The second commit contains the extensions that I have made to it which include:
  -Win32 Implementation
  -Ability to properly open and close more than one serial port (With the REBREQ Union change)
  -Made a few changes to match style of the rest of repository as far as I can tell
## Notes
- For non-posix or win32 systems this will not compile.  I was advised not to add a bunch of ifdefs in lieu of creating stub files for the dev-serial.c when we come across them.  I do not have the ability to test other systems
- I have tested extensively on Linux & Win32 without any issues.  Need someone to test on OS X
## Usage

```
serial-port: open serial://ttyS1/19200
```

Then, read/write binary. Keep in mind it's buffered only by the OS itself, so for long reads, you'll want to do the buffering yourself:

```
response: make binary! 1000
while [not empty? buf: read serial-port][
    append response buf
    wait .2
]
```

### DIFF
--- a/src/core/p-serial.c
+++ b/src/core/p-serial.c
@@ -28,10 +28,10 @@
 ***********************************************************************/
 
 #include "sys-core.h"
-
 #include "reb-net.h"
 #include "reb-evtypes.h"
 
+#define MAX_SERIAL_DEV_PATH 128
 
 /***********************************************************************
 **
@@ -69,10 +69,11 @@
 		switch (action) {
 
 		case A_OPEN:
-			arg = Obj_Value(spec, STD_PORT_SPEC_SERIAL_PATH);
-			req->file.path = VAL_DATA(arg);
+			arg = Obj_Value(spec, STD_PORT_SPEC_SERIAL_PATH);  //Should Obj_Value really return a char* ?
+			req->serial.path = MAKE_STR(MAX_SERIAL_DEV_PATH);
+			TO_OS_STR(req->serial.path, (char *) VAL_DATA(arg), MAX_SERIAL_DEV_PATH);  
 			arg = Obj_Value(spec, STD_PORT_SPEC_SERIAL_SPEED);
-			req->file.size = VAL_INT32(arg); // used for baudrate
+			req->serial.baud = VAL_INT32(arg);
 			//Secure_Port(SYM_SERIAL, ???, path, ser);
 			if (OS_DO_DEVICE(req, RDC_OPEN)) Trap_Port(RE_CANNOT_OPEN, port, -12);
 			SET_OPEN(req);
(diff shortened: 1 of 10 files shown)


src/core/p-serial.c
```diff
@@ -0,0 +1,188 @@
...
+		req->length = len;
+		req->data = VAL_BIN_DATA(spec);
+		req->actual = 0;
+
+		//Print("(write length %d)", len);
+		result = OS_DO_DEVICE(req, RDC_WRITE); // send can happen immediately
+		if (result < 0) Trap_Port(RE_WRITE_ERROR, port, req->error);
+		if (result == DR_DONE) SET_NONE(OFV(port, STD_PORT_DATA));
+		break;
+#if 0
```
hostilefork: Stylistically, I never (myself) use `#if 0`.  One has to scratch one's head to figure out what the "1" condition would be to bring it back.  So I'd say `#if CURE_CODE_1020_IS_RESOLVED` or otherwise a long name expected to be false, which describes the condition under which the code would be enabled.  I like that better than writing long comments.  YMMV.

src/os/posix/dev-serial.c
```diff
@@ -0,0 +1,289 @@
...
+#include <errno.h>
+#include <termios.h>
+
+#include "reb-host.h"
+#include "host-lib.h"
+
+#define MAX_SERIAL_PATH 128
+
+const int speeds[] = {
+	50, B50,
```
hostilefork, line 55: Might be worth mentioning: `/* BXXX constants are defined in termios.h */` - never seen these myself.

src/os/posix/dev-serial.c
```diff
@@ -0,0 +1,289 @@
...
+	return attr;
+}
+
+
+static REBINT Set_Serial_Settings(int ttyfd, int speed)
+{
+	REBINT n;
+	struct termios attr = {};
+
+	// printf("setting attributes: speed %d\n", speed);
```
hostilefork: I'd put these outputs under a DEBUG_SERIAL flag or other construct (or delete them) vs. leaving in commented-out code.

src/os/posix/dev-serial.c
```diff
@@ -0,0 +1,289 @@
...
+**	Local Functions
+**
+***********************************************************************/
+
+static struct termios *Get_Serial_Settings(int ttyfd)
+{
+	struct termios *attr = NULL;
+	attr= (struct termios *) MAKE_NEW(struct termios);
+	if (attr != NULL)
+      	{
```
earl: Formatting here is off. The opening brace should be on the same line as the `if`.

src/os/posix/dev-serial.c
```diff
@@ -0,0 +1,289 @@
...
+        	if (tcgetattr(ttyfd,attr) == -1)
+       		{
```
earl: Formatting

src/os/posix/dev-serial.c
```diff
@@ -0,0 +1,289 @@
...
+          		FREE_MEM(attr);
+          		attr = NULL;
+        	}
+      	}
```
earl, line 91: Indendation for this brace seems off.

kealist:
Found the error on Windows, will post a fix later today

kealist:
Pushed the fix and squashed all my commits into one

Issue #230, opened 2014-12-29
Sequence Point undefined behavior in c-do.c


hostilefork:
[This line in c-do.c](https://github.com/rebol/rebol/blob/master/src/core/c-do.c#L454) is an example of something for which the order of operations is not defined:

```
// Get first block value:
pvs.path = VAL_BLK_DATA(pvs.orig = *path_val);
```

_(Note that VAL_BLK_DATA is a preprocessor macro, and not a function.)_

One might think that means that the compiler could choose to do the `pvs.orig = *path_val`, then call `VAL_BLK_DATA` on that, then put the result in `pvs.path` **or** put the `VAL_BLK_DATA (pvs.orig)` into `pvs.path`, and then do `pvs.orig = *path_val`.  But as pointed out here, it's not an either-or situation in principle.  It's [undefined behavior](http://blog.regehr.org/archives/213).

http://stackoverflow.com/a/13714644/211160

Happened across this just by trying to build the sole file c-do.c with warning levels turned up, adding things like `-Werror=sequence-point`, which would be a good idea in general to do.

hostilefork:
Closing old issues.

PR #231, opened 2014-12-29
Fix sequence point problems

hostilefork:
Proposal from earl is to use this PR as a collection for sequence point issues, first mentioned in #230.

http://stackoverflow.com/questions/4362501/any-good-reason-why-assignment-operator-isnt-a-sequence-point

### DIFF
--- a/src/core/c-do.c
+++ b/src/core/c-do.c
@@ -451,7 +451,8 @@ void Trace_Arg(REBINT num, REBVAL *arg, REBVAL *path)
 	pvs.store = DS_TOP;		// Temp space for constructed results
 
 	// Get first block value:
-	pvs.path = VAL_BLK_DATA(pvs.orig = *path_val);
+	pvs.orig = *path_val;
+	pvs.path = VAL_BLK_DATA(*path_val);
 
 	// Lookup the value of the variable:
 	if (IS_WORD(pvs.path)) {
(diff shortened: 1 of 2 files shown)


earl:
r+

hostilefork:
Before, when doing some other edits, I'd been just kind of eyeballing any assignments on one line and figuring the practice should be avoided...whether they were actual sequence point errors or not.  _(`a = b = 0;` is not, for instance.)_

Having currently built the entire Rebol codebase with strict c99 and c++98 I can now simplify this to just the changes that represent sequence point errors, using the "pervasive approach" I'd mentioned before.  Having trimmed out the stray errors I may update this as a collective patch that fixes the sequence point problems--and _only_ the actual sequence point problems.

One risk of retaining the practice of multiple assignments on one line is that it makes it easier to introduce these errors.  And as mentioned, the heavy use of preprocessor macros can make it "hard to see" that an assignment is not guarded by a function call but actually boils down to a problem.  _But_ if Rebol is altered to be built with high warning levels (which I have)--and these warnings are treated as errors (which I do)--then it becomes unnecessary to police the practice if it is expedient in source, as the compiler can police it before people check it in.

_(In theory.)_

hostilefork:
Fixed in Ren/C as part of [Coherence 1.5](https://github.com/metaeducation/ren-c/pull/12)

PR #232, opened 2014-12-30
Make Halt_State a proper global

hostilefork:
There's is a [comment on Halt_State](https://github.com/rebol/rebol/blob/25033f897b2bd466068d7663563cd3ff64740b94/src/core/c-do.c#L53) which says:

```
static jmp_buf *Halt_State = 0;  //!!!!!!!!!! global?
```

Ten exclamation points and one question mark indicates it's not too surprising that when trying to do some integration with Rebol on a project, this was a problem.  I worked around the problem I had by taking off the `static` and that was enough for what I needed, so I originally submitted this PR for that.  @earl suggested fixing it properly, as he has a better idea of what the multithreading story is.  So I replaced my commit with his.

I tested his change and it worked for my case.  I don't know that this is the kind of thing that's really tested fully by any automated tests... nothing hits Ctrl-C to cancel a running evaluation but direct user interaction.

**Note: This is an important one to integrate for something I'm working on that links with Rebol.**

_(And it means 10 fewer exclamation points and one fewer question mark in the code.)_

### DIFF
--- a/src/core/b-init.c
+++ b/src/core/b-init.c
@@ -899,6 +899,7 @@ static REBCNT Set_Option_Word(REBCHR *str, REBCNT field)
 	PG_Mem_Usage = 0;
 	PG_Mem_Limit = 0;
 	PG_Reb_Stats = Make_Mem(sizeof(*PG_Reb_Stats));
+	Halt_State = 0;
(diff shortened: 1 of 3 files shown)


hostilefork:
Merged into Ren/C, closing.

PR #234, opened 2015-01-25
Add Expand_Frame call to Bind_Block_Words

hostilefork:
I noticed very sketchy behavior of COPY when it was applied to the user context (or really, any object that had been used for binding).

If you want to see the carnage, just try:

```
>> ctx1: bind? 'print
>> ctx2: copy ctx1
>> ctx1/testing: "This 'testing' will appear in ctx2, and it's of type END!"
>> probe ctx2
```

In my brief diagnosis of the issue, I noticed that `append ctx1 'testing` did not have the same problem.  So something about the implementation of appending fields to objects didn't introduce the "copy frailty" that binding had.

What I noticed about other calls to Append_Frame was that they had associated Expand_Frame calls.  I blindly added that call to Bind_Block_Words, and it gave the appearance of correcting the problem.

I submit it as a talking point for whatever the real solution might be.

### DIFF
--- a/src/core/c-frame.c
+++ b/src/core/c-frame.c
@@ -829,7 +829,7 @@
 			else {
 				// Word is not in frame. Add it if option is specified:
 				if ((mode & BIND_ALL) || ((mode & BIND_SET) && (IS_SET_WORD(value)))) {
-					Expand_Frame(obj, 1, 1);
+					Expand_Frame(frame, 1, 1);
 					Append_Frame(frame, value, 0);
 					binds[VAL_WORD_CANON(value)] = VAL_WORD_INDEX(value);
 				}
--- a/src/core/c-frame.c
+++ b/src/core/c-frame.c
@@ -829,6 +829,7 @@
 			else {
 				// Word is not in frame. Add it if option is specified:
 				if ((mode & BIND_ALL) || ((mode & BIND_SET) && (IS_SET_WORD(value)))) {
+					Expand_Frame(obj, 1, 1);
 					Append_Frame(frame, value, 0);
 					binds[VAL_WORD_CANON(value)] = VAL_WORD_INDEX(value);
 				}


earl:
A simpler reproduction:

``` rebol
ctx1: context []
ctx2: copy ctx1
bind/new 'foo ctx1
trim/lines mold/all ctx1  ; == "#[object! [ foo: #[unset!] ]]"
trim/lines mold/all ctx2  ; == "#[object! [ foo: #[end!] ]]"
```

hostilefork:
> That fails to build. Should probably be: Expand_Frame(frame, 1, 1);

Argh.  Step 1: Spend two days tracking down problem.  Step 2: Submit patch on mainline you're not actually building and type in the variable name wrong.  Step 3: Profit?

zsx:
I am a little bit confused about A_APPEND and A_COPY for object:
why does A_APPEND copy the word list, while A_COPY doesn't?

My feeling is that it should be exactly opposite: A_APPEND modifies the words in place, while A_COPY copies both value and words.

Am I missing anything?

earl:
I spent some hours yesterday analysing this. Some findings:

Word list management for objects seems to be intended to be [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write): neither A_MAKE nor A_COPY (which create new objects from existing ones) copy the word list.

Hence, structural mutation of objects has to ensure that the word list is properly copied. As far as I am aware (only from the top of the head, needs some further checking), there are three ways to structurally modify objects: APPEND, EXTEND, BIND/NEW (and BIND/SET, which I'll subsume under BIND/NEW).

APPEND and EXTEND properly copy the word list, BIND/NEW (+ BIND/SET) does not.

This (preliminary) analysis suggests a somewhat different fix to the discussion point submitted by @hostilefork above. Still needs to be converted to a proper fix, though.

gregit:
So, is the following expected result:

> > ctx1: context []
> > == make object! [
> > ]
> > 
> > ctx2: copy ctx1
> > == make object! [
> > ]
> > 
> > append ctx1 'foo
> > == make object! [
> >     foo: unset!
> > ]
> > 
> > ctx2
> > == make object! [
> > ]

hostilefork:
@gregit Seems relatively sensible to me, although I did remark in chat that I thought `ctx1/foo` should give an error vs. the "it's an unset so just be silent".  _(Which may be aggressive about raising errors, but it just seems consistent with typing in foobar in the console and getting an error if it's unset...)_

PR #235, opened 2015-03-02
Pass indexes to ARRAY/initial function

BrianHawley:
When calling ARRAY/initial with a function used to generate the values,
pass the current index at the position the function is being called for,
with each index in a block of sizes provided to create a multidimensional
array being provided as a separate parameter to the function in the order
specified in the sizes block. Uses APPLY so the function can optionally
not take the parameters. Uses an internal function option for efficiency.

See http://issue.cc/r3/2193 for details.


Oldes:
Included in my branch: https://github.com/Oldes/Rebol3/commit/9989527c899df8ec3415dbe61288356ce507c9dc

PR #236, opened 2015-03-03
Set local-lib to refer to the local import library

BrianHawley:
Instead of having the local import library object have no referent once the module is running except the words you bound from it, set the word `local-lib` in the module context to refer to it. This will let the developer control overrides, just like `lib` does globally.

Also set `local-lib` to `system/contexts/user` in the script context. Even though there is nothing for scripts like the object target of local imports in modules, presetting the word will prevent it from being a problem when modules unnecessarily export it, and it will serve as a shortcut way to refer to the context, like `lib`.

See http://issue.cc/r3/2116 for details.

### DIFF
--- a/src/mezz/sys-base.r
+++ b/src/mezz/sys-base.r
@@ -129,6 +129,7 @@ make-module*: func [
 	if find spec/options 'extension [
 		append obj 'lib-base ; specific runtime values MUST BE FIRST
 	]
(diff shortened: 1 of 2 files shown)


earl:
r+

acook:
Wonderfully small changeset for this.

PR #237, opened 2015-04-14
Fix: make-make.r broken for Amiga target

kealist:
Running make-make.r with target 0.1.03 currently errors out in Rebol.   Haven't been able to test the rest of the compilation, but with this change the makefile is generated.


earl:
Looks good, thanks. Also: feel cheered on to get an Amiga build to work :)

earl:
r+

hostilefork:
Merged into Ren/C.

PR #239, opened 2015-04-16
Type-check call of SORT/compare handler

earl:
SORT/compare now checks that the typespec of the comparator function
actually allows calls with values to be compare.

Previously:

```
>> sort/compare [1 2 #[unset!]] :>
;; R3 crashes at this point without a system exception.
```

With this fix:

```
>> sort/compare [1 2 #[unset!]] :>
** Script error: op! does not allow unset! for its value1 argument
** Where: sort
** Near: sort/compare [2 1 unset!] :>
```

This fixes [CureCode issue #1516](http://issue.cc/r3/1516).

### DIFF
--- a/src/core/t-block.c
+++ b/src/core/t-block.c
@@ -382,16 +382,33 @@ static struct {
 
 /***********************************************************************
 **
-*/	static int Compare_Call(const void *v1, const void *v2)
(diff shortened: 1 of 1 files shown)


zsx:
Looks good to me

PR #248, opened 2015-12-14
Avoiding conditional directives that split up statements

RomeroMalaquias:


hostilefork:
Hello @RomeroMalaquias,

The active development branch of Rebol 3 is "Ren-C", literally thousands of commits "in the future":

https://github.com/metaeducation/ren-c

You will find that many things have changed; the posix and linux hosts have been separated, with files factored differently...and the #ifdef you are addressing no longer exists there:

https://github.com/metaeducation/ren-c/tree/master/src/os

It's hard to give a summary of all the changes, the Trello makes some points:

https://trello.com/b/l385BE7a/rebol3-porting-guide-ren-c-branch

It's a work in progress...but the code is locked down much better with a debug build, terminology that's clearer, etc.  And actively vetting and including PRs.  :-)

Chat:

http://rebolsource.net/go/chat-faq

RomeroMalaquias:
Hi @hostilefork , thanks for info.
Can i ask a question? Which style do you prefer? Before or after this commit?

hostilefork:
My absolute preference is not to program in text at all, but using graph-structured editors... and to use language features instead of a preprocessor, and never use token-pasting, etc.  And generally not to program in raw standards-compliant C89.  :-)

But so long as that's the game here, I would say that I don't have any hard and fast rule that says you shouldn't have constructs that split across an #ifdef.  It can save on lines of code here and there.  (And gotos can be helpful too, when coding at this very low level.)

For this project, I say do it only if you're saving several lines of repeated code by doing so.  This wouldn't be a good usage, so I'd likely change it if it were still relevant.

A slight beginning of a style guide for Ren-C is here, though it's just a few cards and I haven't been keeping it up much:

https://trello.com/b/d38MUleH/ren-c-coding-conventions

But if there were more people working on it, there'd be more incentive to do so!  More hands welcome, even on going through and just making it better or cleaner C code.  Or editing comments.  The comments--in fact--define natives inline in the C now:

https://github.com/metaeducation/ren-c/blob/e833e9e2559b72bbefb68180af80589b3e67704d/src/core/n-control.c#L549

And speaking of examples of the kinds of changes that are welcome, that documentation string on handler isn't indented.  So although a commit with just that one change wouldn't be very useful, a full screening of _all_ the natives and making sure they are indented properly, or sweeping through and improving the comments and naming, is the kind of thing that would be great.

PR #251, opened 2018-06-25
Oldes' branch

Oldes:
Just in case someone is interested in other active work than Ren-C branch.
I'm trying to keep it close to the original and so just cherry-picking from here and there and adding some features, which I consider to be good and useful for my own purposes.


src/core/n-control.c
```diff
@@ -254,6 +254,36 @@ enum {
 	return R_ARG1;
 }
 
+/***********************************************************************
+**
+*/	REBNATIVE(protectedq)
```
rebolek, line 249: what does the `q` at the end mean? or is it just a typo?

Oldes: `q` means `question mark` as it is native definition for Rebol's `protected?` function. It's how it was defined in [original sources](https://github.com/rebol/rebol/blob/25033f897b2bd466068d7663563cd3ff64740b94/src/tools/make-boot.r#L136-L149).

